PR #12, opened 2016-10-20
Validate semver in pack.yaml version

emedvedev:
Make `version` in `pack.yaml` strictly conform to semver.

@DoriftoShoes @Kami let me know if you have any objections to this. :)

### DIFF
--- a/.circle/pack.yaml.schema
+++ b/.circle/pack.yaml.schema
@@ -2,18 +2,24 @@ type: object
 properties:
   name:
     type: string
+    description: The pack's primary identifier.
   description:
     type: string
+    description: Brief (1-2 sentences) description.
   keywords:
     type: array
+    description: A list of keywords.
   version:
     type: string
+    description: Version number in the SemVer format (e.g. "0.0.1").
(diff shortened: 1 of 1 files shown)


.circle/pack.yaml.schema
```diff
@@ -8,6 +8,7 @@ properties:
     type: array
   version:
     type: string
+    pattern: "^(?:0|[1-9]\d*)\.(?:0|[1-9]\d*)\.(?:0|[1-9]\d*)(?:-[\da-z\-]+(?:\.[\da-z\-]+)*)?(?:\+[\da-z\-]+(?:\.[\da-z\-]+)*)?$"
```
Kami: Oh, no, it's a regex! 

LGTM, but it would be good to describe valid values in `description` attribute so it's more immediately obvious what are the valid values :)

Kami: Besides that, LGTM

emedvedev: It wouldn't be me if there would be no regexes involved! 😛 Added the description.

Kami:
Looks like YAML doesn't like \ and it needs to be escaped.

```yaml.scanner.ScannerError: while scanning a double-quoted scalar
  in "/home/ubuntu/ci/.circle/pack.yaml.schema", line 15, column 14
found unknown escape character 'd'
  in "/home/ubuntu/ci/.circle/pack.yaml.schema", line 15, column 27
```

Edit: Looks like that just using single instead of double quotes solves the problem :)

PR #13, opened 2016-10-20
Various improvements, index format change

Kami:
This pull request includes some improvements in the index generation code:

1. It always processes files on the disk in the same order and also sorts the resulting JSON data which is written to the file. This ensures that if we provide the same input data, the end result file (final `index.json`) will be the same and won't change.

2. Changes the file format a bit so the packs are stored under the "packs" key and introduces new top-level "metadata" key which includes various index level metadata. Right now it includes the timestamp when the file has been generated and a hash of all the content. Hash won't change if the input data is the same so people who mirror the index can use this attribute to detect changes.

I know this change breaks the current format, but I think it's more future proof and it's better to give us more flexibility by not storing packs as top-level. I know this will also require changes in st2 and index web.

### DIFF
--- a/.circle/deployment
+++ b/.circle/deployment
@@ -11,7 +11,7 @@ git clone https://${MACHINE_USER}:${MACHINE_PASSWORD}@github.com/StackStorm-Exch
 cp pack.yaml ~/index/v1/packs/"${PACK_NAME}".yaml
 
 # Rebuild the index JSON
-~/virtualenv/bin/python ~/ci/.circle/index.py --glob "~/index/v1/packs/*.yaml" --output "~/index/v1/packs/"
+~/virtualenv/bin/python ~/ci/.circle/index.py --glob "~/index/v1/packs/*.yaml" --output "~/index/v1/"
 
 # Check if an icon has been added or changed
 ICON_TARGET="${HOME}/index/v1/icons/${PACK_NAME}.png"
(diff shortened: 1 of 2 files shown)


emedvedev:
Nice! Could be useful in the future. Not just for mirroring, too: we can introduce something new like `suites` at some point, too, and keep the index backwards-compatible.

Kami:
@emedvedev So you are OK with this format? If so, I will go ahead and make necessary changes in web and st2. Is there any other place I should update?

emedvedev:
100% fine. I'll make the changes myself to save some time. :)

Kami:
Sweet, thanks.

Kami:
Pushed another change which allows user to specify input glob and output path. This changes makes it possible to use this script in a single repository deployments (all packs inside a single repo such as st2contrib).

That's how I tested it:

```bash
python .circle/index.py --glob "~/w/stackstorm/st2contrib/packs/*/pack.yaml" --output=~/w/stackstorm-exchange/ci/
```

emedvedev:
Can you share the output? I think there'll be bunch of other issues with single-repo deployments: for one, you won't be able to install from the `repo_url` because it will be auto-generated to point at our Exchange.

Kami:
@emedvedev I didn't actually touch `repo_url` at all, should be fine for now, but yeah we can make some changes to that in the future as well if the need arises.

Kami:
Here are the corresponding changes:

* https://github.com/StackStorm-Exchange/web/pull/4
* https://github.com/StackStorm/st2/pull/2966
* https://github.com/StackStorm-Exchange/index/pull/17

Is this all, or did I miss something?

PR #18, opened 2016-10-29
Store pack content count in index

emedvedev:
Scan for pack content and store the count in the index. Full content metadata is not stored, empty content types are excluded.

### DIFF
--- a/utils/pack_content.py
+++ b/utils/pack_content.py
@@ -1,0 +1,81 @@
+#!/usr/bin/env python
+#
+# A helper script to count content in a pack.
+#
+import os
+import glob
+import yaml
+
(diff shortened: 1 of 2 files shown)


utils/pack_content.py
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/env python
+#
+# A helper script to count content in a pack.
+#
+import os
```
Kami: I think we should depend on st2 (simply clone it, add components to PYTHONPATH) and re-use existing functionality from st2common.

In fact, that's what I need to do anyway for the validation changes (clone st2 and add it to PYTHONPATH for  the validation script).

PR #21, opened 2016-11-24
Various changes to the bootstrap script

Kami:
- [x] Add .gitignore to the repo on bootstrap
- [x] Configure webhook for notifications so we get notifications on changes to our Slack instance


Kami:
I added a change so now when we bootstrap repo we also configure webhook for Slack notifications. This way we don't miss out new changes, comments, etc.

For this to work `SLACK_WEBHOOK_URL` environment variable needs to be set to a valid Slack webhook URL.

PR #28, opened 2016-11-29
Use pack ref instead of pack name for index dictionary key

Kami:
This way it works correctly and as expected with "pack" CLI commands.


emedvedev:
LGTM

PR #36, opened 2017-02-28
Add script which sets pack repo description to match description from pack.yaml

### LINKED ISSUE #35: Make sure all pack repos have a Github repo description set
Right now it looks like that all the pack repositories doesn't have Github repository description set.

I thought exchange bootstrap did that, but maybe I was wrong.

In any case, we should add a small script which takes get token and then sets Github repo description so it matches `description` field from `pack.yaml` metadata file.

After the script is available, we should run it on all the existing pack repos to ensure each repo has a description set.

Having a description set should help with Github and external search and SEO.

Kami:
The title says it all.

In addition to setting Github repo description, it also sets website to https://exchange.stackstorm.org/.

Having description set will make pack discovery through search engines easier.

On a related note - I also wanted to set topics for each repo to match `keywords`, but it looks like there is no API available yet for new Github topics functionality.

Resolves #35.

### DIFF
--- a/tools/set_repo_description.sh
+++ b/tools/set_repo_description.sh
@@ -1,0 +1,44 @@
+#!/usr/bin/env bash
+
+# Script which sets pack Github repo description so it matches "description"
+#
+# Requires: jq
+#
+# attribute from pack.yaml metadata file.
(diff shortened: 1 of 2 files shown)


Kami:
The script run just finished and now all the repos have a description set.

PR #38, opened 2017-05-01
removed broken change to set repo description

LindsayHill:
You can't automatically set the repo description when bootstrapping a new repo. 

The script sets up a new blank repo. You can't then try to pull `pack.yaml` from that repo, before it's had anything merged.


Kami:
Good catch - thanks :+1:

PR #43, opened 2017-05-22
Add a check which verifies pack contains config.schema.yaml and not config.yaml

Kami:
This pull request adds a CI check for each pack which verifies pack contains `config.schema.yaml` file and no `config.yaml` file.

Note: This should only be merged once we update all the existing packs and make sure they all have `config.schema.yaml`.

### DIFF
--- a/.circle/Makefile
+++ b/.circle/Makefile
@@ -72,6 +72,10 @@ packs-tests: requirements .clone_st2_repo .packs-tests
 	. $(VIRTUALENV_DIR)/bin/activate; if [ ! "${CHANGED_YAML}" ]; then echo "No files have changed, skipping run..."; fi; for file in $(CHANGED_YAML); do if [ -n "$$file" ]; then st2-check-validate-yaml-file $$file || exit 1 ; fi; done
 	. $(VIRTUALENV_DIR)/bin/activate; if [ ! "${CHANGED_JSON}" ]; then echo "No files have changed, skipping run..."; fi; for file in $(CHANGED_JSON); do if [ -n "$$file" ]; then st2-check-validate-json-file $$file || exit 1 ; fi; done
 	@echo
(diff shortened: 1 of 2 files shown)


LindsayHill:
👍 Looks good, but as you say, wait until we finish the list here before merging https://github.com/StackStorm/st2/issues/3424#issuecomment-302774214

LindsayHill:
All Exchange packs have now been updated.

But thinking a bit more about this, what about packs that don't have (or need) a `config.schema.yaml`file? e.g. github.com/StackStorm-Exchange/stackstorm-urbandict

Those should not fail validation because they do not contain a config.schema.yaml. They would fail validation if they had an empty config.schema.

Kami:
@LindsayHill Perhaps the easiest way would be to add empty config schema to those packs?

Since we have more packs with config schema than without so I think having a check which checks pack contains a schema is not a bad idea (alternative would be to not have this check at all).

LindsayHill:
Wasn’t there an issue at one stage with having an empty config.schema file? Can’t remember the details, might need to check that. 

But anyway, in general I think that’s an Ok approach. Might also need to update docs to say ‘you must have this file (even if it’s empty)’

LindsayHill:
I don't think this is required anymore. Recent versions now throw ERROR if a pack contains `config.yaml`:

```shell
2018-06-13 16:12:26,666 ERROR [-] Pack "st2" contains a deprecated config.yaml file (/opt/stackstorm/packs/st2/config.yaml). Support for "config.yaml" files has been deprecated in StackStorm v1.6.0 in favor of config.schema.yaml config schema files and config files in /opt/stackstorm/configs/ directory. Support for config.yaml files has been removed in the release (v2.4.0) so please migrate. For more information please refer to https://docs.stackstorm.com/reference/pack_configs.html
```

PR #46, opened 2017-06-23
Enabling test coverage and timing output for pack tests

nmaludy:
I thought it would be a good idea to print out test coverage and timings for unittests in packs similar to what's done in st2 core.

This change depends on st2 PR: https://github.com/StackStorm/st2/pull/3508

And also has supporting documentation here: https://github.com/StackStorm/st2docs/pull/535


Kami:
Now that the corresponding st2 pr has been merged, I will go ahead and merge this.

PR #48, opened 2017-11-28
Save detailed content meta for packs

enykeev:


Kami:
LGTM, but I didn't test it.

PR #59, opened 2018-11-09
Fix Metrics Registration

LindsayHill:
Added step to register metrics drivers. Tested it [here](https://circleci.com/gh/StackStorm-Exchange/stackstorm-mssql/28) with MSSQL pack, pulling this branch of ci repo. That passes OK.

This could be broken out as a separate item in the Makefile. Was just a PoC at this stage. @Kami will leave it to you to decide how best to implement it.

### DIFF
--- a/.circle/Makefile
+++ b/.circle/Makefile
@@ -128,8 +128,8 @@ packs-tests: requirements .clone_st2_repo .packs-tests
 	@echo "================== register metrics drivers ======================"
 	@echo ""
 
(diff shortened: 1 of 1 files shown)


Kami:
LGTM :+1: 

I will also look into pinning Jinja2 dependency to a specific version in st2 fixed-requirements.txt to avoid caching issues (good practice to do it anyway, we just need to make sure it's in sync with Orquesta so there won't be any conflicts).

PR #60, opened 2018-11-10
Reduce noise during "Run Tests" Circle CI build step

Kami:
This pull request improves the output generated during "Dependencies" and "Run Tests" steps which run on each Circle CI build.

We have two main steps inside our pack build workflow - one to install the dependencies and the other one to run all the various (lint) checks and pack tests.

Before this change we actually installed dependencies in two places - inside "dependencies" step and also inside "run tests" step.

This meant that "Run Tests" step produced a lot of dependency installation related output which made actual "valuable" lint check and test output hard to spot.

And not to mention that it was confusing that the dependencies were installed in multiple places.

I added a new Make target and left "all" target as is, because the idea is that you can also use this ``Makefile`` with your project so it works out of the box outside our Circle CI build environment.

Before: https://circleci.com/gh/StackStorm-Exchange/stackstorm-mssql/28
After: https://circleci.com/gh/StackStorm-Exchange/stackstorm-mssql/36

### DIFF
--- a/.circle/Makefile
+++ b/.circle/Makefile
@@ -19,6 +19,9 @@ COMPONENTS_RUNNERS := $(wildcard /tmp/st2/contrib/runners/*)
 .PHONY: all
 all: requirements lint packs-resource-register packs-tests
 
+.PHONY: all-ci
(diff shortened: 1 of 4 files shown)


LindsayHill:
Looks good, makes things cleaner

PR #61, opened 2018-12-14
Add support for Python 3.x checks to pack Circle CI config

Kami:
This pull request adds support for running various lint checks and pack tests under Python 3.6 (in addition to Python 2.7).

In addition to new Python 3 checks, I also added new "compile" check for Python 2.7 and Python 3.6 job which verifies Python syntax in pack files is valid.

A couple of things to keep in mind:

1. For the time being, we will run those checks on all the packs, but won't consider Python 3.x check and test failures as fatal (aka failing the build).
This will allow us to identify "problematic" packs. The goal is to add new ``python_version`` attribute to ``pack.yaml`` file and utilize that attribute in CI to determine if pack supports Python 3. For packs which claim support for Python 3, build will fail if Python 3 checks fail, and for ones which don't, it wont (will open a PR for that in StackStorm/st2 shortly).
2. Needed to make small change to script in ``st2sdk`` because CI scripts use StackStorm from git checkout and all the shebangs in scripts in the repo explicitly default to python2.7 binary.

## TODO

- [x] Publish new version of st2sdk to PyPi
- [x] Fix CI in st2sdk
- [ ] Update all packs to use this new Circle CI config

### DIFF
--- a/.circle/Makefile
+++ b/.circle/Makefile
@@ -20,7 +20,7 @@ COMPONENTS_RUNNERS := $(wildcard /tmp/st2/contrib/runners/*)
 all: requirements lint packs-resource-register packs-tests
 
 .PHONY: all-ci
-all-ci: .flake8 .pylint .copy-pack-to-subdirectory .configs-check .metadata-check .packs-resource-register .packs-tests
+all-ci: .compile .flake8 .pylint .copy-pack-to-subdirectory .configs-check .metadata-check .packs-resource-register .packs-tests
 
 .PHONY: lint
 lint: requirements flake8 pylint configs-check metadata-check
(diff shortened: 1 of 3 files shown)


LindsayHill:
I'm generally happy with the approach. I like the idea of running Py3 checks, but not failing on them. Not at this stage, anyway. Just means we'll have to go and manually eyeball it to check Py3 status.

Will leave it until next year before it becomes a breaking change

Kami:
I will go ahead, merge this and add new CircleCI config to all the pack repos.

Once this PR is finished and merged (https://github.com/StackStorm/st2/pull/4474) I will open a new PR which fails Python 3 build for packs which declare support for Python 3.

Kami:
It's on - https://circleci.com/gh/StackStorm-Exchange.

I'm working on finishing and merging https://github.com/StackStorm/st2/pull/4474. Once that's done, I will go ahead and update pack metadata for all the Exchange packs and CI as mentioned above.

PR #64, opened 2019-01-22
Run nightly build for every pack

Kami:
This pull request updates pack Circle CI config so we run nightly build (minus the deploy step) for each pack.

### Background

While working on exchange Pack Python 3 support, it turned out that a lot of pack builds were broken for one reason or another.

There were various reasons for that (dependency was updated, token has expired, etc.), but the root cause was that there were no commits / pushes to that pack repo so we didn't have any builds running which would detect that early.

With this change we will now detect such issues early and also be able to fix them early.

### DIFF
--- a/.circle/circle.yml.sample
+++ b/.circle/circle.yml.sample
@@ -112,7 +112,8 @@ jobs:
 
 workflows:
   version: 2
-  build_test_deploy:
+  # Workflow which runs on each push
+  build_test_deploy_on_push:
     jobs:
       - build_and_test_python27
       - build_and_test_python36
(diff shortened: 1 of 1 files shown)


Kami:
Confirmed the cron syntax is working here - https://circleci.com/gh/StackStorm-Exchange/stackstorm-csv

Kami:
Per discussion with @bigmstone and @LindsayHill on Slack I pushed a change so we will run the build weekly every Saturday at 1 AM UTC instead of every day to avoid hitting Circle CI limits for open source repos.

PR #68, opened 2019-02-05
Bump pyyaml version

LindsayHill:


Kami:
Pushes a small change (b824c67) which shouldn't be needed with this upgrade anymore, but still.

LindsayHill:
Ta

LindsayHill:
@Kami looks like https://github.com/StackStorm-Exchange/ci/commit/b824c6761dea90e14bfdc1ee35d81979d7c9529c causes problems - see https://circleci.com/gh/StackStorm-Exchange/stackstorm-icinga2/61

I'll leave that to you to decide how you want to fix it

Kami:
Sorry about that - I simply reverted that change.

With new version of pyyaml ``load`` now works as ``safe_load`` so we should be fine.

PR #70, opened 2019-03-25
Upgrade lint dependency versions so the match versions in StackStorm/st2 repo

Kami:
This pull request updates lint dependency versions (flake8, pylint) so they match versions used in StackStorm/st2 repo.

Those new versions might identify some new issues on new (weekly) CI runs which may need to be fixed for the builds to pass again.


Kami:
@warrenvw Good catch, will do.

PR #72, opened 2019-09-10
Include available versions for all the packs in the index

Kami:
This pull request updates generate index script so we now also include all the available versions for each pack in the index.json file.

## Implementation

This implementation utilizes Github API to retrieve all the available tags for a particular pack repo.

## TODO

- [x] Handle Github authentication so we get a higher rate limit

### DIFF
--- a/.circle/index.py
+++ b/.circle/index.py
@@ -9,10 +9,14 @@ import hashlib
 from glob import glob
 from collections import OrderedDict
 
+import requests
+
 from st2common.util.pack import get_pack_ref_from_metadata
 
 EXCHANGE_NAME = "StackStorm-Exchange"
 EXCHANGE_PREFIX = "stackstorm"
+
+SESSION = requests.Session()
 
 
 def build_index(path_glob, output_path):
@@ -44,6 +48,8 @@ def build_index(path_glob, output_path):
             EXCHANGE_NAME, EXCHANGE_PREFIX, sanitized_pack_name
         )
 
+        pack_meta['tags'] = get_tags_for_pack(pack_ref)
+
         # Note: Key in the index dictionary is ref and not a name
         result['packs'][pack_ref] = pack_meta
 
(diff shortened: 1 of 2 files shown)


Kami:
I updated the script so we use Github credentials we generate for each pack repo.

I first thought we might have API rate limiting issues, but we use different Github credentials for each pack deployment so we should be fine.

The whole process is still a bit inefficient since I followed the existing approach which means we rebuild the whole index on each pack deployment. I will file this under "feature improvement / optimization" :)

I also confirmed it's working at https://github.com/StackStorm-Exchange/stackstorm-slack/pull/52 (https://circleci.com/gh/StackStorm-Exchange/stackstorm-slack/255).

Kami:
@armab I agree, ``versions`` / ``available_versions`` is a better name for that attribute (and it's also consistent with the existing ``version`` attribute).

At this point no functionality relies on this attribute yet, but there are various use cases for it (e.g. right now listing available pack versions in StackStorm CLI depends on git, etc.).

Kami:
@armab Here we go - 8b86314c7db89598213b4d6bc305d7861c6beb47 (https://github.com/StackStorm-Exchange/index/commit/1a165bd0f517caf123b48a82ae5759d8d8ef06b8).

.circle/index.py
```diff
@@ -67,6 +80,32 @@ def build_index(path_glob, output_path):
...
+    Retrieve all the available versions for a particular pack.
+
+    NOTE: This function uses Github API.
+    """
+    url = ('https://api.github.com/repos/%s/%s-%s/tags' %
+           (EXCHANGE_NAME, EXCHANGE_PREFIX, pack_ref))
+    resp = SESSION.get(url)
+
+    if resp.status_code != 200:
+        print('Got non 200 response: %s' % (resp.text))
```
blag: Would it be helpful to let the user know what status code was received? All they know here is that it wasn't a 200.

.circle/index.py
```diff
@@ -67,6 +80,32 @@ def build_index(path_glob, output_path):
...
+        return None
+
+    versions = []
```
blag: Why not a list comprehension here?

blag: `.get(..., None)` is explicit, but redundant, since `.get()` returns `None` by default if the key doesn't exist.

Also, there is a potential bug here.

```python
item.get('name', None)
```

so you are protecting against the case where an `item` doesn't have a `name` key, which is good. But, if `name` doesn't exist, `.get()` returns `None`, which then has `startswith()` run on it. This will result in (I think) an `AttributeError`:

```
>>> a = {
...     'key': 'value'
... }
>>> a.get('name', None).startswith('v')
  File "<stdin>", line 1, in <module>
AttributeError: 'NoneType' object has no attribute 'startswith'
```

This logic is much better represented in a list comprehension:

```python
return list(reversed(sorted([
    item['name'].replace('v', '')
    for item in resp.json()
    if item.get('name', '').startswith('v')
])))
```

That fixes the bug and reads better IMO.

.circle/index.py
Kami: I can also include the status code, but the error message itself should already be helpful enough.

.circle/index.py
Kami: Just a style preference, I like to use list comprehensions for simple stuff, this would likely result in two or more line long list comprehension, where, imo, regular for loop is a better fit.

.circle/index.py
```diff
@@ -67,6 +80,32 @@ def build_index(path_glob, output_path):
...
+
+    for item in resp.json():
+        if item.get('name', None).startswith('v'):
```
blag:
```suggestion
        if item.get('name', '').startswith('v'):
```

Kami: Good catch :+1:

PR #73, opened 2019-09-11
Fix index bug

Kami:
This pull request fixes a bug inadvertently introduced in #72.

Index incorrectly started picking up old versions - https://github.com/StackStorm-Exchange/index/commit/f56625926604c37cd50cad6f19c1c91cf82dfa49.

The bug was related to the code re-org I did. I reorganized the code so we create version tags before we create the index.

As part of the tag creating process, we check out every tag (from newest to oldest). To fix the issue, we need to switch back to the latest commit (master) after processing the tag.

### DIFF
--- a/.circle/deployment
+++ b/.circle/deployment
@@ -54,6 +54,10 @@ done
 # Afer pushing the version, make sure we check out the latest commit
 # This is important so index reflects latest changes
 git checkout master
+git reset --hard master
+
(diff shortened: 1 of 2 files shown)


Kami:
Confirmed it's working - https://github.com/StackStorm-Exchange/index/commit/100cda2fd6d64a1922188d02262e76fc417da244.

I will go ahead and merge it.

PR #75, opened 2019-09-13
[SECURITY] Use PyYAML's yaml.safe_load when loading YAML files

blag:
I immediately merged this PR just in case there is a CircleCI config somewhere that is vulnerable.

This fixes a security flaw where PyYAML could unsafely read a YAML file. So a malicious user _may_ have been able to craft a malicious YAML file, commit it, push it, and create a pull request. Then the CircleCI task would have unsafely loaded the malicious file. This could have happened on **any** pack on StackStorm-Exchange.

With this PR, the YAML file is now safely loaded.

### DIFF
--- a/.circle/index.py
+++ b/.circle/index.py
@@ -40,7 +40,7 @@ def build_index(path_glob, output_path):
     counter = 0
     for filename in generator:
         with open(filename, 'r') as pack:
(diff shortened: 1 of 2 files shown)


Kami:
:+1:

PR #77, opened 2019-09-13
Default to subclassing yaml.SafeLoader

blag:
`yaml.safe_load` doesn't accept a `loader` argument, but we should default to loading YAML safely.

Luckily, everywhere `ordered_load` was used already passed in `yaml.SafeLoader` to subclass, so this shouldn't break anything. This PR does allow users of `ordered_load` to use non-safe loader classes, but now all users of that function have to explicitly enable the unsafe behavior and safe loading is the default.

Tested with the stackstorm-git pack.

### DIFF
--- a/utils/pack_content.py
+++ b/utils/pack_content.py
@@ -45,7 +45,7 @@ RESOURCE_LOCATOR = {
 }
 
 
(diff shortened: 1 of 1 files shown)


blag:
Merging this without review so it's in place for the weekend CI runs.

PR #78, opened 2019-09-14
Force check

blag:
Fix assignment of `FORCE_CHECK_ALL_FILES` variables, also do better at checking for empty and non-empty strings.

And finally, fix the assignments to the `GIT_BRANCH` and `PYTHON_VERSION` variables. The `GIT_BRANCH` assignment was broken because it displayed all of the available branches (with the active branch first). And the `PYTHON_VERSION` assignment was broken because `python --version` prints the Python version to stderr, not stdout, so `PYTHON_VERSION` was always empty.


blag:
Also merging this in time for the weekend CI runs.

blag:
This was tested with the stackstorm-aws pack.

PR #79, opened 2019-09-21
Convert Makefile to use Python's Invoke project

blag:
This PR converts almost all of the targets in the `Makefile` into Invoke tasks. The `Makefile` is kept and vastly simplified into just setting up the environment for invoke, then passing all targets to it. Invoke handles everything else.

This PR isn't strictly 100% necessary, but since the `Makefile` was fairly simple and straightforward it was a good target to illustrate how a `Makefile` can be broken up into multiple different modules and run with Invoke with full reverse compatibility for any scripts that still call `make`. Some of the scripts in this repo do call make with this `Makefile`, and those were intentionally left alone as a sentinel for reverse compatibility. This PR is a good baby step towards [replacing the (very complex/convoluted) `Makefile` in st2](https://github.com/StackStorm/st2/pull/4782), since this `Makefile` was 100% composed of targets marked with `.PHONY`, so it was just a centralized way to run scripts and didn't actually build anything or cache any build artifacts, which is the entire purpose of make to begin with.

I also simplified some of the behavior in `.circle/dependencies`, for the sake of sanity:

1. The `requirements` and `requirements-ci` targets both had the exact same recipe (verified with diff), so I also changed the call to `make requirements-ci` to just `make requirements`.
2. There was a repeated pattern in the `Makefile`:
   ```make
   .PHONY:
   install-runners: .install-runners  # non-dotted target is just redirected to the dotted target

   .PHONY:
   .install-runners:  # the dotted target actually did the work
       # actually do stuff here
   ```
   and usually the dotted target was not called directly. However, there were a few direct calls to the `.install-runners` (dotted) target, so I replaced them to just call the `install-runners` (non-dotted) target.

I explicitly relatively import the task modules since implicit relative imports have been phased out in Python 3. Do note that in `tasks/utils.py`, there is still an `import git` - this is not an implicit relative import of a tasks module, it is an absolute import for the GitPython module in the virtualenv.

Tested with the AWS pack:

* [Python 2.7](https://circleci.com/gh/StackStorm-Exchange/stackstorm-aws/411)
* [Python 3.6](https://circleci.com/gh/StackStorm-Exchange/stackstorm-aws/410)

Please review this thoroughly, since this is used for **all** pack CI runs, including the 130+ runs over the weekend.

### DIFF
--- a/.circle/Makefile
+++ b/.circle/Makefile
@@ -1,249 +1,33 @@
 ROOT_DIR ?= $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
-ifndef FORCE_CHECK_ALL_FILES
-CHANGED_FILES := $(shell $(CI_DIR)/utils/git-changes files)
-CHANGED_PY := $(shell ${CI_DIR}/utils/git-changes py)
-CHANGED_YAML := $(shell $(CI_DIR)/utils/git-changes yaml)
-CHANGED_JSON := $(shell $(CI_DIR)/utils/git-changes json)
-endif
-CHANGED_DIRECTORIES := $(shell $(CI_DIR)/utils/git-changes directories)
 VIRTUALENV_DIR ?= virtualenv
 ST2_REPO_PATH ?= /tmp/st2
 ST2_REPO_BRANCH ?= master
-FORCE_CHECK_ALL_FILES ?= false
-FORCE_CHECK_PACK ?= false
 
-export ST2_REPO_PATH ROOT_DIR FORCE_CHECK_ALL_FILES FORCE_CHECK_PACK
+export ST2_REPO_PATH ST2_REPO_BRANCH ROOT_DIR
 
-# All components are prefixed by st2
-COMPONENTS := $(wildcard /tmp/st2/st2*)
-COMPONENTS_RUNNERS := $(wildcard /tmp/st2/contrib/runners/*)
-
-.PHONY: all
-all: requirements lint packs-resource-register packs-tests
-
-.PHONY: all-ci
-all-ci: compile .flake8 .pylint .copy-pack-to-subdirectory .configs-check .metadata-check .packs-resource-register .packs-tests
-
-.PHONY: lint
-lint: requirements flake8 pylint configs-check metadata-check
-
-.PHONY: flake8
-flake8: requirements .flake8
-
-.PHONY: pylint
-pylint: requirements .clone_st2_repo .pylint
-
-.PHONY: configs-check
-configs-check: requirements .clone_st2_repo .copy-pack-to-subdirectory .configs-check
-
-.PHONY: metadata-check
-metadata-check: requirements .metadata-check
-
-# Task which copies pack to temporary sub-directory so we can use old-style check scripts which
-# # require pack to be in a sub-directory
-.PHONY: .copy-pack-to-subdirectory
-.copy-pack-to-subdirectory:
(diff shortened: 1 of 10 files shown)


tasks/check.py
```diff
@@ -0,0 +1,88 @@
...
+from invoke import run, task
+
+from . import copy
+
+
+@task
+def compile_(ctx):
+    print("======================= compile ========================")
+    print("------- Compile all .py files (syntax check test) ------")
+    compileall.compile_dir(".", rx=re.compile(r"/virtualenv|virtualenv-osx|virtualenv-py3|.tox|.git|.venv-st2devbox"), quiet=True)
```
Kami: Will this correctly throw and exit with non zero if compile_dir fails?

tasks/tests.py
```diff
@@ -0,0 +1,61 @@
...
+    print("==================== packs-resource-register ====================")
+    print("")
+    base_branch = os.environ.get('BASE_BRANCH', 'origin/master')
+    files = []
+    top_level_git_dir = ctx.run('git rev-parse --show-toplevel').stdout.splitlines()[0]
+    with ctx.cd(top_level_git_dir):
+        files = ctx.run("git diff --relative --diff-filter=ACMRTUXB "
+                        "--name-only {}".format(base_branch)).stdout.splitlines()
+    if files:
+        ctx.run("st2-check-print-pack-tests-coverage {pack_dir}".format(pack_dir=top_level_git_dir))
```
Kami: I assume this should be calling a different script which performs the resource register check?

.circle/Makefile
```diff
@@ -1,249 +1,33 @@
...
-.PHONY: virtualenv
 virtualenv: $(VIRTUALENV_DIR)/bin/activate
 $(VIRTUALENV_DIR)/bin/activate:
 	@echo
 	@echo "==================== virtualenv ===================="
 	@echo
 	test -d $(VIRTUALENV_DIR) || virtualenv --no-site-packages $(VIRTUALENV_DIR)
+
+$(VIRTUALENV_DIR)/bin/invoke: $(VIRTUALENV_DIR)
+	. $(VIRTUALENV_DIR)/bin/activate && pip install invoke
```
Kami: Do we want to pin a version?

Kami:
I'm fine with that change. Definitely much better than convoluted Makefile.

As mentioned in my other PR, it perhaps wouldn't be a bad idea to document some of that behavior, etc. since in theory, users could also use this Makefile for CI for their custom packs.

tasks/check.py
Kami: EDIT:

> Recursively descend the directory tree named by dir, compiling all .py files along the way. Return a true value if all the files compiled successfully, and a false value otherwise.

https://docs.python.org/3/library/compileall.html#compileall.compile_dir

So it looks like we may want to check the return value and throw / exit with non zero if this function returns False?

.circle/Makefile
blag: I don't think so, since that would immediately become a maintenance burden. Invoke was born out of the Fabric project, so the developers have had a good amount of time to figure out what works and what doesn't. I think that means that the API is pretty stable and won't really change a whole lot from here on out. And we aren't really using any in-depth parts of Invoke's API, so our code is very likely to remain compatible with newer versions of Invoke.

tasks/tests.py
blag: Good catch, will fix.

nmaludy:
I'm good with this too! Much easier to debug than Make.

+1

PR #81, opened 2020-02-01
Force Python 3 test failures to fail the build

blag:


punkrokk:
Am I wrong or should this really require a script to create a PR in each pack repo to resolve broken py3 tests? Is it reasonable to expect the next person that happens to do a PR to have to fix the tests?

nmaludy:
Yeah we really need tooling to help with that.

In Puppet land we use either modulesync or PDK. Basically both tools have a 'template" repo they base off of, then when you run the sync it clones your main repo and syncs with the template.

Modulesync, I believe, also had the ability to make changes over a big list of repos. Think of this like cookiecutter on steroids.

https://github.com/voxpupuli/modulesync
https://github.com/voxpupuli/modulesync_config

https://github.com/puppetlabs/pdk
https://github.com/puppetlabs/pdk-templates

I'm sure we could use these tools. Not sure if writing something ourselves is worthwhile?

blag:
You're not wrong, but I'm the one working on fixing all packs that don't work on Python 3. It's a long story, but I was going to merge this in before our weekly pack CI runs (which test every Python file in the pack, and not just changed Python files like the PR tests do) this weekend, so I could see packs actually fail. And I plan on reverting this on Monday.

The way we have it setup now, the Python 3 PRs can pass, but once you merge into the pack `master` branch, _that_ CI run fails. :rage: See my recent [AWS pack PR](https://github.com/StackStorm-Exchange/stackstorm-aws/pull/92) ([test results](https://github.com/StackStorm-Exchange/stackstorm-aws/runs/416288967))for an example of that ([test results after merge to `master`](https://circleci.com/gh/StackStorm-Exchange/stackstorm-aws/523)).

blag:
Tested this with the AWS pack, results here:
https://circleci.com/gh/StackStorm-Exchange/stackstorm-aws/530

I'm going to merge this in for the weekend pack CI runs, and then revert it on Monday.

blag:
I should also add that for pack PRs this weekend, the CircleCI builds will only fail (due to this change) if they touch a module that is not Python 3 compliant and do not fix the module themselves (because CI runs for PRs only check files that are changed in the PR itself, not all files).

Since most of our pack PRs happen on weekdays, and our CI for PRs only checks changed files, I think the impact on other users will be minimal this weekend.

punkrokk:
I do like the idea of running them with weekly then reverting to have a hit list for remediation. 

Cheers,

JP

PR #86, opened 2020-04-16
Make sure we run unit tests if FORCE_CHECK_ALL_FILES is true

Kami:
My PR here (https://github.com/StackStorm-Exchange/stackstorm-circle_ci/pull/8#issuecomment-614640449) seems to have uncovered a bug with unit tests not being ran when ``FORCE_CHECK_ALL_FILES`` is set (e.g. for master builds or weekly builds).

I have no idea how long has this been broken / not working correctly (could have been since the beginning).

This being broken basically means we didn't run any pack unit tests as part of our weekly builds (that's obviously bad and it means we likely missed a bunch of failures).

### DIFF
--- a/.circle/Makefile
+++ b/.circle/Makefile
@@ -181,10 +181,10 @@ compile:
 	@echo "==================== packs-resource-register ===================="
 	@echo
 	. $(VIRTUALENV_DIR)/bin/activate; \
-	if [ ${NUM_CHANGED_FILES} -eq 0 ]; then \
-		echo "No files have changed, skipping run..."; \
-	else \
+	if [ "$${FORCE_CHECK_ALL_FILES}" = "true" ] || [ ${NUM_CHANGED_FILES} -gt 0 ]; then \
 		st2-check-register-pack-resources /tmp/packs/$(PACK_NAME) || exit 1; \
+	else \
+		echo "No files have changed, skipping run..."; \
 	fi;
 
 .PHONY: .packs-tests
@@ -204,10 +204,10 @@ compile:
 	@echo
 	@echo "==================== pack-missing-tests ===================="
 	@echo
-	if [ ${NUM_CHANGED_FILES} -eq 0 ]; then \
-		echo "No files have changed, skipping run..."; \
-	else \
+	if [ "$${FORCE_CHECK_ALL_FILES}" = "true" ] || [ ${NUM_CHANGED_FILES} -gt 0 ]; then \
 		st2-check-print-pack-tests-coverage $(ROOT_DIR) || exit 1; \
+	else \
+		echo "No files have changed, skipping run..."; \
 	fi;
 
 # Target which veries repo root contains LICENSE file with ASF 2.0 content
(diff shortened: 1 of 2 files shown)


Kami:
^ On that note, we need to check other targets as well and make sure we run them on all the files when ``FORCE_CHECK_ALL_FILES`` is true.

armab:
Yes, I think @blag was touching a lot of this Exchange CI machinery recently. Should have more context.

blag:
I believe that this was broken way back in September in [this commit](https://github.6b8573f3650c5f42e7f455a20077088ea26a893f).

Before that change, the `FORCE_CHECK_ALL_FILES` and `FORCE_CHECK_PACK` make variables would default to `? false`:

```diff
 VIRTUALENV_DIR ?= virtualenv
 ST2_REPO_PATH ?= /tmp/st2
 ST2_REPO_BRANCH ?= master
-FORCE_CHECK_ALL_FILES =? false
-FORCE_CHECK_PACK =? false
+FORCE_CHECK_ALL_FILES ?= false
+FORCE_CHECK_PACK ?= false
```

which _was_ okay because our comparisons weren't very explicit:

```diff
-if [ "$${FORCE_CHECK_ALL_FILES}" ]; then \
+if [ "$${FORCE_CHECK_ALL_FILES}" = "true" ]; then \
```

so those tests actually got run no matter what.

I'll figure out where all I need to add `[ "$${FORCE_CHECK_ALL_FILES}" = "true" ]` checks, push them to this branch, and get that all merged in before our weekend pack CI, and then spot check a few of those to make sure they run ALL the tests.

Thanks for catching this @kami!

Kami:
Thanks for taking this across the finish like and testing and confirming the fix.

blag:
I have tested this with the AWS pack by running a CircleCI job with SSH, and then manually setting `FORCE_CHECK_ALL_FILES` to `true`, `false`, and empty. This should be RFR.

.circle/Makefile
```diff
@@ -69,6 +67,8 @@ compile:
 	@echo "======================= compile ========================"
 	@echo "------- Compile all .py files (syntax check test) ------"
 	if python -c 'import compileall,re; compileall.compile_dir(".", rx=re.compile(r"/virtualenv|virtualenv-osx|virtualenv-py3|.tox|.git|.venv-st2devbox"), quiet=True)' | grep .; then exit 1; else exit 0; fi
+	@echo
+	@echo "All files compiled successfully"
```
Kami, line 71: There is exit 0 above and the command doesn't run in a subshell so I assume this output will never be printed, but I could be wrong.

blag: Oh, I think you're right.

blag: Nope, and I figured out why. Each line of a `Makefile` is run in its own Bash shell, that's why exporting variables across lines of a make recipe don't work, or activating a virtualenv in one line and using it in another.

So if the `if python -c` line fails, `make` will catch it and stop. If it doesn't, then that line exits with status `0`, and the echo lines are run in their own Bash shells.

PR #87, opened 2020-04-25
Check for the MIT license for the napalm_logs and netbox packs

blag:
Unfortunately the napalm_logs and netbox packs are not licensed under the Apache 2.0 License like the rest of the packs are.

I initiated [a PR](https://github.com/StackStorm-Exchange/stackstorm-netbox/pulls/17) to change the license, but we cannot do so without the consent of every contributor to those packs, and at this point I don't think we are going to get that from all contributors. Same story with the napalm_logs pack.

So, instead of checking for the Apache 2.0 license on those packs and failing every weekly CI run, we simply hardcode those two packs to exempt them from the Apache 2.0 License requirement. Luckily both are licensed under the very liberal MIT license, and were initially authored by John Anderson, so for those two packs we check for the MIT license instead.

I manually tested this with both packs, and the AWS pack to ensure it doesn't break anything.

Since the Bash commands to actually run the checks are getting very complex, and because the error messages for all of the checks are very well written and make it very clear how to fix the issue, I muted them from being printed before execution with the `@`.

In my opinion, the ancient Unix philosophy of "be quiet unless you can't do your job" does not apply to tests, where you want to verify that the tests either successfully ran or successfully failed. Therefore, I have each branch each print out different success message indicating the detected license along with a smiley face, and I tweaked the error message if the MIT license is not found.

Fixes:
* https://circleci.com/gh/StackStorm-Exchange/stackstorm-napalm_logs/166
* https://circleci.com/gh/StackStorm-Exchange/stackstorm-netbox/254
* https://circleci.com/gh/StackStorm-Exchange/stackstorm-netbox/255

### DIFF
--- a/.circle/Makefile
+++ b/.circle/Makefile
@@ -216,24 +216,36 @@ compile:
 	fi;
 
 # Target which veries repo root contains LICENSE file with ASF 2.0 content
+# We allow the napalm_logs and netbox packs to have MIT licenses, since they
+# adopted those licenses before this check was put in place, so they are
+# effectively grandfathered in
+# However, all other packs, including any new packs, are required to be
+# licensed under the Apache 2.0 License
 .PHONY: .license-check
 .license-check:
 	@echo
 	@echo "==================== license-check ===================="
 	@echo
-	if [ ${NUM_CHANGED_FILES} -eq 0 ] && [ "$${FORCE_CHECK_ALL_FILES}" = "false" ]; then \
+	@if [ ${NUM_CHANGED_FILES} -eq 0 ] && [ "$${FORCE_CHECK_ALL_FILES}" = "false" ]; then \
 		echo "No files have changed, skipping run..."; \
 	else \
 		if [ ! -f "$(ROOT_DIR)/LICENSE" ]; then \
 			echo "Missing LICENSE file in $(ROOT_DIR)"; \
 			exit 2;\
 		fi;\
-		cat $(ROOT_DIR)/LICENSE | grep -q "Apache License"  || (echo "LICENSE file doesn't contain Apache 2.0 license text" ; exit 2); \
-		cat $(ROOT_DIR)/LICENSE | grep -q "Version 2.0"  || (echo "LICENSE file doesn't contain Apache 2.0 license text" ; exit 2); \
-		cat $(ROOT_DIR)/LICENSE | grep -q "www.apache.org/licenses/LICENSE-2.0"  || (echo "LICENSE file doesn't contain Apache 2.0 license text" ; exit 2); \
+		if [ "$$(python -c 'import yaml; f = open("/home/circleci/repo/pack.yaml"); print(yaml.safe_load(f.read())["name"]); f.close();')" = "napalm_logs" ] || \
+		   [ "$$(python -c 'import yaml; f = open("/home/circleci/repo/pack.yaml"); print(yaml.safe_load(f.read())["name"]); f.close();')" = "netbox" ]; then \
+			cat $(ROOT_DIR)/LICENSE | grep -q "MIT License"  || (echo "LICENSE file doesn't contain MIT license text" ; exit 2); \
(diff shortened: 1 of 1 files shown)


armab:
@nmaludy Awesome idea about license in `pack.yaml` metadata!
Can you create a feature request in StackStorm/st2 for that?

blag:
I'm 👎 on supporting license declarations in `pack.yaml`.

Having solved licensing issues in the past, I don't like the idea of declaring a `license` in `pack.yaml` because (1) the use case for it doesn't fit the Exchange, and (2) it also completely duplicates where a license is declared.

1. PyPI packages can use trove classifiers to declare their license so that PyPI itself can implement package search and let searchers restrict their results to only packages licensed under specific licenses. This is especially useful, for instance, if your organization has restrictions on what license third party packages can use. As it stands, right now, StackStorm itself is licensed under the Apache 2 license, and **every** pack in Exchange is also licensed under the Apache 2.0 license (except for the two exceptions this is grandfathering in). Note that that exception is not likely to cause a problem for users, because the MIT license has fewer restrictions than the Apache license (specifically around patent grants) so if you can install StackStorm, you are currently guaranteed to be able to install any pack in Exchange, without any further licensing considerations. This is important in organizations with "hyperactive" legal departments who want to review every license in every context. Since Kami added this check to the `ci` repo, every future pack submitted to Exchange must be licensed under the Apache license, which maintains this guarantee. If we want to allow packs to be licensed under something other than Apache 2.0 (ignoring these two exceptions), that would break this guarantee, and that's a much larger change than supporting a key in `pack.yaml` and should therefore be written up and discussed in an official ST2 proposal.
2. Declaring the license in `pack.yaml` sounds like it would make these checks easier, but it would actually make it an order of magnitude more difficult. I have found a few Python packages, used by StackStorm, that have declared one license in their `setup.py` but had an entirely different license in the actual `LICENSE` file. In that case, which one takes precendence? What if a user installed a pack that was marked as `Apache 2.0` in `pack.yaml`, but the actual `LICENSE` was `All your base are now belong to us`, or something proprietary, or something legally nefarious? The license "You have zero legal right to use this pack until you pay me one million Bitcoin for each action you run or trigger you send" is a perfectly legal license, but it is obviously not something any sane person would agree to. Currently, our CI process checks for the existing Apache 2.0 license text (and currently fails on these two exceptional packs), but if we allow packs to use non-Apache licenses, not only will this complicate pack installation for our users, it will also complicate the CI checks. We would then have to check the pack's `LICENSE` file for the text of allowed licenses (who decides that?), _also_ check `pack.yaml` for a license declaration, and then _also_ check that both licenses agree.

I don't see the benefit in declaring the pack license in the `pack.yaml` file. At best it provides no benefit for us without much larger (and controversial) organizational changes, and at worst it vastly complicates our CI checks, our pack submission/acceptance criteria for acceptable licenses, and the pack installation for our users.

Right now, pack submission/acceptance criteria regarding pack licensing is simple, straightforward, and exactly matches StackStorm itself: new packs must all be licensed under the Apache 2.0 license, and all current packs must maintain their Apache 2.0 license (except for these two annoying exceptions that have very little hope of ever being relicensed to Apache 2.0 under US copyright law). Checking pack licenses is already complex enough, and we only have two exceptional packs, and luckily they both use the same exceptional license. Supporting anything more complicated than all of this should have distinct, observable, and justifiable benefits, and go through a much more rigorous vetting process.

nmaludy:
@blag Don't disagree with your thoughts there.

Was just an observation from the way it works in the Puppet community where we declare our license as part of the `metadata.json`  (think `pack.yaml`) and then it's validated as part of CI (guessing with a regex, honestly haven't looked).

I don't have a strong opinion either way, just saw what one community did and it was successful, thought we could potentially steal that idea for us.

I'm good either way!

armab:
I see the license metadata in absolutely every package manager and that makes sense to me.

@nmaludy I don't think this should be part of this discussion or PR however.
Can you please create a feature request in https://github.com/stackstorm/st2 for that?

PR #89, opened 2020-07-10
Fix pylint by locking isort module below version 5

bgastinne-algolia:
With the new release of isort 5 that gets installed as a dependency of pylint, the pylint CI step is currently broken with the following error:
```
AttributeError: module 'isort' has no attribute 'SortImports'
```
This dependency issue is only fixed in newer pylint versions (>=2.4)


nmaludy:
For anyone else that is here, looks like we pin to 1.9.4 because pylint 2.x is python 3 only.

Issue #94, opened 2020-08-05
Action tests fail if you want to use common libs


punkrokk:
If you are using the st2.conf setting: 

```
[packs]
enable_common_libs = True
```

Then circle tests fail for actions since the python path is: 
```
Running pylint on pack: repo
PYTHONPATH=/tmp/st2/st2actions:/tmp/st2/st2api:/tmp/st2/st2auth:/tmp/st2/st2client:/tmp/st2/st2common:/tmp/st2/st2debug:/tmp/st2/st2exporter:/tmp/st2/st2reactor:/tmp/st2/st2stream:/tmp/st2/st2tests:/home/circleci/repo/actions/lib:/tmp/st2/st2common:
PYTHON_BINARY=/home/circleci/virtualenv/bin/python
Using config file /home/circleci/ci/lint-configs/python/.pylintrc-exchange
```

I think the solution is to be able to pass an envvar to the build job for a pack that detects that it is using common libs then loads an [st2_test.conf](https://github.com/StackStorm-Exchange/ci/blob/master/.circle/test#L41-L42) file that sets the `enable_common_libs` option

PR #95, opened 2020-08-05
Common libs

punkrokk:
Fixes: #94 

### Improvements:
* add env var to support packs using [common_libs](https://docs.stackstorm.com/reference/sharing_code_sensors_actions.html) option
* Inject the needed PATH for the common_libs folder: `export PYTHONPATH=/home/circleci/repo/lib:${PYTHONPATH}`
* Use a st2.conf file that has the `packs. enable_common_libs = True` so the st2 registration tests run properly.

If you wish to use this, in circleci for the pack, (or defined in the `.circleci/config.yml` file) you need to set the environment variable `COMMON_LIBS=true`


punkrokk:
I could see false positives if something is in the wrong place or it’s disabled on install. 

But for testing - I guess it should be ok? 

JP Bourget / @punkrokk

PR #98, opened 2020-10-26
Improve deployment scripts

blag:
When packs on Exchange fail to deploy, we don't have any good ways of detecting that failure and exiting gracefully with an informative or useful error message. On top of that, we don't point TSC members to the PAT tools in this `ci` repo, nor do we mention that the required credentials are in LastPass. And then on top of that, the reset token script tries to silently read environment variables.

This PR improves this entire situation, by "closing the loop" and laying down a trail of breadcrumbs for uninitiated users, both community members and TSC members.

First, it asks the user for the GitHub username and password, and then asks for the CircleCI token. If any of those are specified in environment variables, it silently uses those and skips asking the user for them. When the script asks for the values, it notes where those values can be found (eg: the specific accounts and fields in LastPass) to hopefully help the user determine whether or not they have access to the required GitHub and CircleCI credentials.

Second, it isn't clear that the pack name should be specified _without_ the exchange prefix (`stackstorm`, as in `stackstorm-vault`). Instead of throwing an error, the script now skips adding the prefix itself if the name already has it.

Next, instead of writing the GitHub PAT to a temporary file and reading it in again, the script now just echoes the returned PAT to a variable and interpolates that variable when setting it in CircleCI.

Finally, the index generating script now counts how many repositories it cannot clone and process, points the user to the reset tool in `StackStorm-Exchange/ci`, and mentions that a TSC member with the appropriate permissions may need to run the reset tool.

### DIFF
--- a/.circle/index.py
+++ b/.circle/index.py
@@ -40,6 +40,7 @@ def build_index(path_glob, output_path):
     generator = sorted(glob(path_glob))
 
     counter = 0
+    failed_count = 0
     for filename in generator:
         with open(filename, 'r') as pack:
             pack_meta = yaml.safe_load(pack)
@@ -55,6 +56,9 @@ def build_index(path_glob, output_path):
         )
 
         versions = get_available_versions_for_pack(pack_ref)
+
+        if versions is None:
+            failed_count += 1
 
         if versions is not None:
             pack_meta['versions'] = versions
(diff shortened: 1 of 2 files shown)


nmaludy (approved):
Awesome! Thank you for this!

PR #104, opened 2021-02-06
Use Github graphql api to avoid rate limiting

cognifloyd:
This adjusts the index regeneration so that it retrieves the list of all
pack versions (repo tags) in as few API calls as possible.

At 1 call per pack per pack deployment, multiple pack deployments can
easily hit the rate limit. Using graphql reduces the number of calls
that count toward the rate limit (today) to 2 per pack deployment.
Tags are apparently a very inexpensive API request, so github is only
counting 1 call per page. More complex queries can count as more than
one call per page, so this is very fortunate for our use case.

Graphql limits each page to 100 results, so with about 160 packs, that
is only 2 pages. So, we won't hit a 3rd page until we get over 200
packs.

### DIFF
--- a/.circle/deployment
+++ b/.circle/deployment
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -e
+set -ex
 
 FORCE_REBUILD_INDEX="${FORCE_REBUILD_INDEX:-0}"
 
--- a/.circle/deployment
+++ b/.circle/deployment
@@ -18,8 +18,8 @@ then
   exit 1
 fi
 
-# make `gh` available for github API calls
-~/ci/.circle/install_gh
+# TODO: remove the install_gh call once circleci config is updated in all packs
+type gh &>/dev/null || ~/ci/.circle/install_gh
 
 # TODO: figure out how to make deploy.py rebuild the index.
 # python ~/packs/.circle/deploy.py pack.yaml "${CIRCLE_PROJECT_REPONAME}"
--- a/.circle/deployment
+++ b/.circle/deployment
@@ -23,7 +23,7 @@ fi
 
 # Clean up so the script can be retried in case of failure (e.g. race)
 rm -rf ~/index
-git clone https://${MACHINE_USER}:${MACHINE_PASSWORD}@github.com/StackStorm-Exchange/index ~/index 2>/dev/null
+git clone -q -- "https://${MACHINE_USER}:${MACHINE_PASSWORD}@github.com/StackStorm-Exchange/index" ~/index
 
 echo "Processing pack ${PACK_NAME}"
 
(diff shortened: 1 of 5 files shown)


cognifloyd:
I've tested my changes manually. Both `get_available_versions` and `get_available_versions_for_pack` work.
The only way to test this E2E is to merge it and then deploy a pack update that triggers index regeneration.

cognifloyd:
I've tested the `gh` install block and modified it so it works.
I haven't tested E2E yet, as my tests are failing due to CircleCI issues.

.circle/deployment
```diff
@@ -18,6 +18,9 @@ then
   exit 1
 fi
 
+# make `gh` available for github API calls
+~/ci/.circle/install_gh
```
cognifloyd, line 22: This could be installed during the `dependencies` phase instead. The only gotcha is that it requires updating `.circleci/config.yml` for all packs (Why I'm doing now anyway, so that would not be a big deal).

.circle/index.py
```diff
@@ -105,27 +123,87 @@ def build_index(path_glob, output_path):
...
+            } for repo in repos if repo["name"].startswith(EXCHANGE_PREFIX)
+        ]
+        for pack in packs:
+            pack_name = pack["reponame"][prefix_len:]
+            if pack_name not in PACK_VERSIONS:
+                PACK_VERSIONS[pack_name] = []
+            PACK_VERSIONS[pack_name].extend(
+                tag.replace("v", "") for tag in pack["tags"]
+                if tag.startswith("v")  # only version tags
+            )
```
blag, line 180: This is a great place to use `dict.setdefault`:

```suggestion
            PACK_VERSIONS.setdefault(pack_name, []).extend(
                tag.replace("v", "") for tag in pack["tags"]
                if tag.startswith("v")  # only version tags
            )
```

Example:

```python
>>> d = {'a': [1], 'b': [2]}
>>> d.setdefault('a', []).extend([3, 4, 5])
>>> d.setdefault('c', []).extend([6, 7, 8])
>>> d
{'a': [1, 3, 4, 5], 'c': [6, 7, 8], 'b': [2]}
```

blag (changes requested):
One small change. Otherwise looks good.

blag (approved):

armab:
The new CI error messages we're getting for Exchange Packs after this PR:

```
+ /home/circleci/virtualenv/bin/python /home/circleci/ci/utils/pack_content.py --input . --output /home/circleci/index/v1/packs/jmx
++ git -C /home/circleci/index status -s
+ [[ -n  M v1/packs/jmx/pack.yaml ]]
+ [[ 0 == \1 ]]
+ /home/circleci/virtualenv/bin/python /home/circleci/ci/.circle/index.py --glob '~/index/v1/packs/*/pack.yaml' --output '~/index/v1/'
gh: Bad credentials (HTTP 401)
  File "/home/circleci/ci/.circle/index.py", line 217, in <module>
    get_available_versions()
  File "/home/circleci/ci/.circle/index.py", line 166, in get_available_versions
    repos = page["data"]["repositoryOwner"]["repositories"]["nodes"]
KeyError: 'data'
```

@blag Is that because of expired creds?

@cognifloyd I guess we need a bit better error reporting there.

blag:
Yes it is. And yes we do.

blag:
Captured in #109.

PR #105, opened 2021-02-09
Use PATs weekly to prevent expiration

cognifloyd:
PATs expire when unused for a year: https://github.community/t/api-access-token-expiry/14522

So, we need to use it regularly. Adding an API hit to the weekly test run should keep the PAT active indefinitely.

Here we use the dependencies script so that it runs as early as possible each week before most other things might fail and kill the build.

Also, we use `|| true` so that any PR runs can continue passing even if the PAT has expired.


blag:
1000000x yes.

blag (approved):

PR #107, opened 2021-02-09
Improve usability of PAT regen script

cognifloyd:
Improve the output while running the script.
Do not request the github password as it is no longer needed.

### DIFF
--- a/tools/reset_github_user_token_and_update_circleci.sh
+++ b/tools/reset_github_user_token_and_update_circleci.sh
@@ -65,7 +65,7 @@ end tell" | osascript
         echo "PR welcome to add support for your Browser."
         exit 1
     fi
-if [[ "${OS}" == "Linux" ]]; then
+elif [[ "${OS}" == "Linux" ]]; then
     # We're on linux
     # Gnome, KDE, and other desktops each have different commands to run *.desktop files.
     # So, we roll our own runner (based on: https://askubuntu.com/a/664272 )
--- a/tools/reset_github_user_token_and_update_circleci.sh
+++ b/tools/reset_github_user_token_and_update_circleci.sh
@@ -9,6 +9,25 @@
 # * USERNAME: a GitHub user to run the script under (Exchange bot).
 # * CIRCLECI_TOKEN: a CircleCI token for the Exchange organization.
 #
+# HISTORICAL NOTE: GitHub disabled basic auth *with account passwords*[1]. Before that change,
+# this script automatically generated the Personal Access Tokens (PATs) via an API call with
+# account-password based basic auth. GitHub probably intends to force a "human in the loop" for
+# PAT generation. As such, this script was modified to open a private/incognito browser window
+# with the token details pre-filled. This is now the expected process for using this script:
+#
+#  1. Open https://github.com/settings/tokens/new in a new private/incognito browser window.
+#  2. Log in with stackstorm-neptr, the StackStorm-Exchange's bot account (not a personal account).
+#  3. Run this script:
+#         USERNAME=stackstorm-neptr CIRCLECI_TOKEN=... \
+#         ./tools/reset_github_user_token_and_update_circleci.sh pack1 pack2 pack3 pack4
+#  4. The script prints instructions and opens the token generation page in the default browser
+#  5. Manually scroll down and hit "Generate Token" (PAT name and scope are already filled in).
+#  6. Copy PAT contents/value and paste in the terminal window (it is read as a secure password).
+#  7. Wait for the script to load the PAT into the MACHINE_PASSWORD var in CircleCI
+#     (The CircleCI APIs can be quite slow, so doing many packs takes awhile).
+#  8. The script loops through steps 4-8 until all packs have been processed.
+#
+# [1] https://developer.github.com/changes/2020-02-14-deprecating-password-auth/
 
 if [[ ! $# -gt 0 ]];
 then
--- a/tools/reset_github_user_token_and_update_circleci.sh
+++ b/tools/reset_github_user_token_and_update_circleci.sh
@@ -135,12 +135,9 @@ for PACK in $@; do
 
     # If we're doing more than one pack, add some whitespace and a counter
     if [[ $NUM_PACKS -gt 1 ]]; then
-        # Space things out a bit
-        echo
-        echo
-
-        # Use printf instead of echo since it doesn't include a newline terminator
-        printf "[${PACK_NUM}/${NUM_PACKS}]"
+        # Space things out a bit with \n\n
+        # And use printf instead of echo since so it add another newline
+        printf "\n\n[${PACK_NUM}/${NUM_PACKS}]"
     fi
     echo " ${REPO_NAME}"
     echo
--- a/tools/reset_github_user_token_and_update_circleci.sh
+++ b/tools/reset_github_user_token_and_update_circleci.sh
@@ -7,7 +7,6 @@
 #
 # The following env variables must be specified:
 # * USERNAME: a GitHub user to run the script under (Exchange bot).
-# * PASSWORD: password for the user (not a token).
 # * CIRCLECI_TOKEN: a CircleCI token for the Exchange organization.
 #
 
@@ -43,14 +42,6 @@ then
     echo "Default: $DEFAULT_USERNAME (just hit enter to use this)"
     read USERNAME
     USERNAME="${USERNAME:-$DEFAULT_USERNAME}"
-fi
-
-if [[ -z "$PASSWORD" ]];
-then
-    echo "What is the password for the GitHub user (${USERNAME})?"
-    echo "This password is stored in LastPass under the ${DEFAULT_USERNAME}"
-    echo "account."
-    read -s PASSWORD
 fi
 
 if [[ -z "$CIRCLECI_TOKEN" ]];
--- a/tools/reset_github_user_token_and_update_circleci.sh
+++ b/tools/reset_github_user_token_and_update_circleci.sh
@@ -16,8 +16,31 @@ then
     exit 1
 fi
 
-BROWSER_NAME=Firefox
-BROWSER="/Applications/Firefox.app/Contents/MacOS/firefox -private-window"
+if [[ "$(uname)" == "Darwin" ]]; then
+    # We're on macOS
+    # Figure out the default browser
+    DEFAULT_BROWSER=$(plutil -p ~/Library/Preferences/com.apple.LaunchServices/com.apple.launchservices.secure.plist | grep 'https' -b3 |awk 'NR==3 {split($4, arr, "\""); print arr[2]}')
+    if [[ "$DEFAULT_BROWSER" == "org.mozilla.firefox" ]]; then
+        BROWSER_NAME=Firefox
+        BROWSER_COMMAND="/Applications/Firefox.app/Contents/MacOS/firefox -private-window"
+    elif [[ "$DEFAULT_BROWSER" == "com.google.chrome" ]]; then
+        BROWSER_NAME=Chrome
+        BROWSER_COMMAND='/Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome --incognito'
+    elif [[ "$DEFAULT_BROWSER" == "com.apple.safari" ]]; then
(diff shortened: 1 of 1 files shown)


blag (approved):

tools/reset_github_user_token_and_update_circleci.sh
```diff
@@ -45,14 +60,6 @@ then
...
 fi
 
-if [[ -z "$PASSWORD" ]];
-then
-    echo "What is the password for the GitHub user (${USERNAME})?"
-    echo "This password is stored in LastPass under the ${DEFAULT_USERNAME}"
-    echo "account."
-    read -s PASSWORD
-fi
-
```
armab, line 55: Wait a minute, at which point Github Password is not used anymore? What's used instead then and where it's documented?

tools/reset_github_user_token_and_update_circleci.sh
```diff
@@ -17,7 +16,7 @@ then
     exit 1
 fi
 
-BROWSER_NAME=firefox
+BROWSER_NAME=Firefox
 BROWSER="/Applications/Firefox.app/Contents/MacOS/firefox -private-window"
```
armab, line 20: ??

How Firefox and MacOS path ended up in automation script like that?

tools/reset_github_user_token_and_update_circleci.sh
```diff
@@ -7,7 +7,6 @@
 #
 # The following env variables must be specified:
 # * USERNAME: a GitHub user to run the script under (Exchange bot).
-# * PASSWORD: password for the user (not a token).
 # * CIRCLECI_TOKEN: a CircleCI token for the Exchange organization.
```
armab, line 10: Is Github Username and CircleCI token enough now for running this script?
what kind of auth is used?

tools/reset_github_user_token_and_update_circleci.sh
```diff
@@ -64,35 +71,58 @@ fi
...
+    SCOPE=$(repo_scope ${PACK})
+
+    # If we're doing more than one pack, add some whitespace and a counter
+    if [[ $NUM_PACKS -gt 1 ]]; then
+        # Space things out a bit
+        echo
+        echo
+
+        # Use printf instead of echo since it doesn't include a newline terminator
+        printf "[${PACK_NUM}/${NUM_PACKS}]"
```
armab, line 90:
```suggestion
        # Space things out a bit
        echo
        echo

        # Use printf instead of echo since it doesn't include a newline terminator
        printf "[${PACK_NUM}/${NUM_PACKS}]"
```

is there a better construction than this?

tools/reset_github_user_token_and_update_circleci.sh
```diff
@@ -64,35 +71,58 @@ fi
...
     # CircleCI: specify the credentials (the machine login and the new user-scope token)
     echo "CircleCI: Setting credentials (machine login and user-scoped token)"
-    curl -sS --fail -X POST --header "Content-Type: application/json" \
-        -d '{"name":"MACHINE_USER", "value":"'"${USERNAME}"'"}' \
-        "https://circleci.com/api/v1.1/project/github/${EXCHANGE_ORG}/${REPO_NAME}/envvar?circle-token=${CIRCLECI_TOKEN}"
+    # CircleCI's API is sloooooow, so minimize the number of calls we make to it
+    # TODO: Hide this behind a command line flag
+    # curl -sS --fail -X POST --header "Content-Type: application/json" \
+    #     -d '{"name":"MACHINE_USER", "value":"'"${USERNAME}"'"}' \
+    #     "https://circleci.com/api/v1.1/project/github/${EXCHANGE_ORG}/${REPO_NAME}/envvar?circle-token=${CIRCLECI_TOKEN}"
```
armab, line 121: Why this line was commented out and why it's not removed?

armab (changes requested):
I'm confused by this PR. It advertizes improved usability, but includes hacks and unclear what exactly it does and why.

How's the GH Password is not needed anymore and why it was replaced with the token, just because someone got the token only?

How we ended up with the mac os firefox string there?

It feels like it does something hacky for specific environment and use case, instead of providing toolset for the whole team and following the established processes.

tools/reset_github_user_token_and_update_circleci.sh
cognifloyd: As of #106, `$PASSWORD` is no longer used in the script, so there is no need to prompt for it.

#106 was necessary because [Github disabled basic auth with user/password](https://developer.github.com/changes/2020-02-14-deprecating-password-auth/), so we cannot use it to automatically generate the PATs any more. [This line](https://github.com/StackStorm-Exchange/ci/pull/106/files#diff-5b060d9c9235f7c266766f99962be0fa99e168790c367228eb4ba3e30e2fe0b5L59) is where we used to use `$PASSWORD`.

tools/reset_github_user_token_and_update_circleci.sh
cognifloyd: @blag is the only person that uses this script. As such, it was designed for his setup using Firefox on Mac. I put these in vars so that we might be able to prompt for which browser to use at some future point, if anyone else on the TSC volunteers to help @blag maintain this part of the Exchange.

tools/reset_github_user_token_and_update_circleci.sh
cognifloyd: Addressed in the other comment about PASSWORD.

tools/reset_github_user_token_and_update_circleci.sh
cognifloyd: I don't understand your suggestion. Could you edit/rephrase?

tools/reset_github_user_token_and_update_circleci.sh
cognifloyd: We still need it for new packs.

As the TODO says, we need to allow enabling it with a command line flag like `--add-machine-user`. However as we (well @blag) were only working with existing packs, we commented this out to speed up fixing all the existing packs by avoiding one SLOW CircleCI api call per pack.

tools/reset_github_user_token_and_update_circleci.sh
cognifloyd: To clarify further: There are no APIs that we can use to generate PATs. Someone has to manually generate the token in the Github web ui. However, we can prefill the form and open a browser window to save as many steps as possible in this manual process.
We use private/incognito browsing because the tokens need to be linked to a bot account that is a member of the StackStorm-Exchange organization, not a personal account. Private/Incognito simplifies keeping that account separate from a personal account.

blag: Fixed this.

tools/reset_github_user_token_and_update_circleci.sh
blag: This script doesn't need to be perfect to be useful.

blag:
The "established processes" don't work anymore because GitHub disabled basic authentication _with account passwords_. You can still use basic auth, but you have to use a PAT as the password. And PATs have scopes applied to them, so they are intentionally limited and restricted from doing certain things - like adding more PATs.

I believe GitHub's intent is to force a "human in the loop" when it comes to generating PATs. Luckily GitHub has made it easier to autofill the PAT name and check different PAT scopes, so we open a URL with HTTP GET parameters and GitHub autofills the rest for us.

So, as I have recently been that "human in the loop" for 80+ packs, the user flow is this:

1. Open a private/incognito browser window and sign in as stackstorm-neptr (the private/incognito mode is to separate it away from my normal browser window, where I am logged into it with my @blag account).
2. Browse to the account settings > Developer Settings > Personal Access Tokens.
3. Put in the stackstorm-netpr password again to activate GitHub's "sudo mode".
4. Run the script:
   ```bash
   USERNAME=stackstorm-neptr CIRCLECI_TOKEN=... ./tools/reset_github_user_token_and_update_circleci.sh pack1 pack2 pack3 pack4
   ```
5. The script prints out what it's doing, and opens a new private/incognito tab to the generate PAT page on GitHub.
6. The PAT name is already filled in, and the `public_repo` scope is already checked, so all I have to do is scroll down and click the "Generate Token" button.
7. Copy the PAT contents/value and paste it into my terminal where the script is running (the script securely reads the PAT like a password, so it's not printed when it is pasted).
8. The script takes the PAT and sets the `MACHINE_PASSWORD` environment variable value to the PAT contents for the pack's CircleCI project (this takes a few seconds, which quickly adds up when you're trying to get through 80+ packs at once).
9. The script goes back to step 5 until all packs are processed.

It's an alright solution. It's not great, I would rather have this completely automated, especially because PATs expire after one year of "non use", but it's a lot easier and more accurate than filling in those fields and checking those boxes by hand.

All of the hacks in this script are working around quirks in GitHub, CircleCI, or Bash. I've tried to document them where I can. It's not perfect, but it's a drastic improvement over having to do all of this manually.

Hopefully, with the other changes @cognifloyd has made, we won't need to constantly refresh tokens every weekend.

cognifloyd:
There we go, @blag and I added ~50 lines of browser detection logic for Mac + Linux. It should be a bit more generic now if others want to use it.

tools/reset_github_user_token_and_update_circleci.sh
armab: At least it's possible to do
```bash
printf "\n\n[${PACK_NUM}/${NUM_PACKS}]"
```

armab:
So the Github behavior has changed significantly. That makes sense, thanks for more info! :+1: 
If 3rd Maintainer comes to a PR and the context is hidden that much, - it's a sign of a problem.
We sometimes work in pairs, but it shouldn't result in something like https://github.com/StackStorm/st2-auth-ldap/pull/83.

I understand a lot of work was done via private conversations and I'd encourage to collaborate about the operational things in public channels like `#exchange` (following #releasemgmt practices) as well as keep Github Issues and PRs better documented for the history and keep up with the existing level of quality.

Every piece of code here is designed to make every Maintainer use it and I know @nmaludy is also involved in the Exchange work. Any TSC member or newcomer should be able to navigate through this without guessing if the script was ever finished.

tools/reset_github_user_token_and_update_circleci.sh
armab: That makes sense.
Can we include that info really briefly somewhere in the script header?

armab:
This also might be helpful: https://github.com/orgs/StackStorm-Exchange/projects/1

tools/reset_github_user_token_and_update_circleci.sh
cognifloyd: Fixed

cognifloyd:
I added a header comment with usage instructions / historical note on why it is opening a browser window.
Plus, I dropped the no-scope token generation.

tools/reset_github_user_token_and_update_circleci.sh
cognifloyd: Fixed

cognifloyd:
I think all raised issues have been resolved. Please merge.

tools/reset_github_user_token_and_update_circleci.sh
```diff
@@ -7,18 +7,104 @@
...
+    esac
+done
 
 if [[ ! $# -gt 0 ]];
 then
-    echo "Usage: $0 <pack> [<pack> <pack> ...]"
+    echo "Usage: $0 [--set-user] <pack> [<pack> <pack> ...]"
+    echo
+    echo "By default, only sets MACHINE_PASSWORD to minimize CircleCI API calls."
+    echo "Use --set-user if you need to also configure MACHINE_USER in CircleCI."
```
armab, line 51: :+1:  including this in Usage.
It's unclear though why someone would need `MACHINE_USER` in CircleCI. Providing a better explanation here would be helpful for someone using it or understand when to use it. 

> Use --set-user if you need to also configure MACHINE_USER in CircleCI to be able to do <...>.

armab (approved):
LGTM

Left one comment to address.

tools/reset_github_user_token_and_update_circleci.sh
cognifloyd: fixed. Please merge

PR #108, opened 2021-02-10
[BUGFIX] Fix support for non-master default branch in packs

### LINKED ISSUE #109: Improve error handling and error message during deploy
Follow up to #104.

When the `index.py` script does not have the correct credentials, it does not properly detect this and proceeds to error out later on when it attempts to extract information from the response:

```
+ /home/circleci/virtualenv/bin/python /home/circleci/ci/.circle/index.py --glob '~/index/v1/packs/*/pack.yaml' --output '~/index/v1/'
gh: Bad credentials (HTTP 401)
  File "/home/circleci/ci/.circle/index.py", line 217, in <module>
    get_available_versions()
  File "/home/circleci/ci/.circle/index.py", line 166, in get_available_versions
    repos = page["data"]["repositoryOwner"]["repositories"]["nodes"]
KeyError: 'data'
```

The script should better detect when it doesn't have the appropriate credentials and give a more user friendly error message about what the underlying issue likely is, and most importantly what the user can do to fix it (eg: go regenerate the PAT for the pack, then set the `MACHINE_PASSWORD` environment variable in the CircleCI project settings, and rerun the CircleCI job).

Hopefully the PATs will expire less often now that we "use" them in the weekend CI runs, but that means that we will not be as familiar with those errors when they do arise since they will happen less frequently, so detecting errors and being more explicit about the exact problem and the way to fix it will help point ~us~ future maintainers to the solution in the future.

cognifloyd:
- Fix support for non-master default pack branches

    If any packs end up using `main` instead of `master` as their default
    branch, then deploying the index will fail without this change.

    Currently only the `test` and `test2` packs use an non-master default
    branch (`aaa`). So, the impact of this change is minimal.

- Provide PAT error message as early as possible

    Resolves #109 by printing a helpful error message much sooner if a PAT has expired.

- Add `DO_NOT_ADD_TO_EXCHANGE_INDEX` for CI packs

    This variable should be set to `true` (a string, not a boolean)
    in CircleCI's web UI for each test pack. This way, such test packs
    can be used to test everything in the CI workflows including
    changes to the `deploy` step's `deployment` script.

### DIFF
--- a/.circle/deployment
+++ b/.circle/deployment
@@ -21,6 +21,10 @@ fi
 # TODO: remove the install_gh call once circleci config is updated in all packs
 type gh &>/dev/null || ~/ci/.circle/install_gh
 
+# GH_TOKEN is used by gh which is used in functions.sh
+export GH_TOKEN=${MACHINE_PASSWORD}
+source ~/ci/tools/functions.sh
+
 # TODO: figure out how to make deploy.py rebuild the index.
 # python ~/packs/.circle/deploy.py pack.yaml "${CIRCLE_PROJECT_REPONAME}"
 
@@ -54,10 +58,11 @@ do
   fi
 done
 
+DEFAULT_BRANCH=$(_gh_default_branch)
 # Afer pushing the version, make sure we check out the latest commit
 # This is important so index reflects latest changes
-git checkout master
-git reset --hard master
+git checkout ${DEFAULT_BRANCH}
+git reset --hard ${DEFAULT_BRANCH}
 
 # Verify there are no unstaged changes
 git status
--- a/.circle/deployment
+++ b/.circle/deployment
@@ -110,10 +110,11 @@ if ! git -C ~/index diff --quiet --exit-code --cached
 then
   echo "Updating the index repo..."
   git -C ~/index commit -m "Update the ${PACK_NAME} pack."
-  if [[ ${SKIP_INDEX:-false} == "true" ]]; then
-    echo "Skipping index update due to SKIP_INDEX=${SKIP_INDEX}"
-    echo "SKIP_INDEX is for packs that are only used in CI;"
-    echo "Normal packs push to the index repo at this point."
+  if [[ ${DO_NOT_ADD_TO_EXCHANGE_INDEX:-false} == "true" ]]; then
+    echo "Normal packs update the Exchange index at this point."
+    echo "This pack will not be added to the Exchange index because of this env var:"
+    echo "DO_NOT_ADD_TO_EXCHANGE_INDEX=${DO_NOT_ADD_TO_EXCHANGE_INDEX}"
+    echo "(This var should only be set per-repo in the CI system's UI.)"
   else
     git -C ~/index push -q origin
     echo "Index updated."
--- a/.circle/deployment
+++ b/.circle/deployment
@@ -27,7 +27,7 @@ source ~/ci/tools/functions.sh
 
 DEFAULT_BRANCH=$(_gh_default_branch)
 # Give a good error message after the first use of github API in this script.
-if $? != 0; then
+if [[ $? != 0 ]]; then
     echo "Error retrieving data with github graphql API."
     echo "The GitHub PAT might need to be regenerated:"
     echo "https://github.com/settings/tokens/new?scopes=public_repo&description=CircleCI%3A%20stackstorm-${PACK_NAME}"
--- a/.circle/deployment
+++ b/.circle/deployment
@@ -110,8 +110,14 @@ if ! git -C ~/index diff --quiet --exit-code --cached
 then
   echo "Updating the index repo..."
   git -C ~/index commit -m "Update the ${PACK_NAME} pack."
-  git -C ~/index push -q origin
-  echo "Index updated."
+  if [[ ${SKIP_INDEX:-false} == "true" ]]; then
+    echo "Skipping index update due to SKIP_INDEX=${SKIP_INDEX}"
+    echo "SKIP_INDEX is for packs that are only used in CI;"
+    echo "Normal packs push to the index repo at this point."
+  else
+    git -C ~/index push -q origin
+    echo "Index updated."
+  fi
 
   # echo "Updating the repo description..."
   # PACK_DESCRIPTION=$(cat ~/index/v1/index.json | jq ".packs.${PACK_NAME}.description")
--- a/.circle/deployment
+++ b/.circle/deployment
@@ -25,6 +25,16 @@ type gh &>/dev/null || ~/ci/.circle/install_gh
 export GH_TOKEN=${MACHINE_PASSWORD}
 source ~/ci/tools/functions.sh
 
+DEFAULT_BRANCH=$(_gh_default_branch)
+# Give a good error message after the first use of github API in this script.
+if $? != 0; then
+    echo "Error retrieving data with github graphql API."
+    echo "The GitHub PAT might need to be regenerated:"
+    echo "https://github.com/settings/tokens/new?scopes=public_repo&description=CircleCI%3A%20stackstorm-${PACK_NAME}"
+else
+    echo "GitHub PAT is active."
+fi
+
 # TODO: figure out how to make deploy.py rebuild the index.
 # python ~/packs/.circle/deploy.py pack.yaml "${CIRCLE_PROJECT_REPONAME}"
 
@@ -58,7 +68,6 @@ do
   fi
 done
 
-DEFAULT_BRANCH=$(_gh_default_branch)
 # Afer pushing the version, make sure we check out the latest commit
 # This is important so index reflects latest changes
 git checkout ${DEFAULT_BRANCH}
--- a/.circle/index.py
+++ b/.circle/index.py
@@ -155,7 +155,7 @@ def get_available_versions():
     if proc.returncode != 0:
         sys.exit(
             "Error retrieving data with github graphql API.\n"
-            "The Github PAT might need to be regenerated:\n"
+            "The GitHub PAT might need to be regenerated:\n"
             "https://github.com/settings/tokens/new?scopes=public_repo"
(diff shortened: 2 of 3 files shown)


armab (changes requested):
You noticed that there are 2 repositories affected:
* https://github.com/StackStorm-Exchange/stackstorm-test/
* https://github.com/StackStorm-Exchange/stackstorm-test2/

But we never did nor should publish them to Exchange Index.

Looking at the commit log history for those repositories, sounds like the `deploy` step should be removed or adjusted instead as no-deploy.

cognifloyd:
I added `SKIP_INDEX` to address preventing a pack from being added to the index.

See StackStorm-Exchange/stackstorm-test#16 and StackStorm-Exchange/stackstorm-test2#8

armab:
Left comments in both PRs.

If we never use non-master branches in Exchange repos for deployment, - this functionality is not needed.
Would be great to avoid overloading the scripts and stick to previous easiest solutions in those repos.

cognifloyd:
We should probably also remove [`python3_test`](https://github.com/StackStorm-Exchange/stackstorm-python3_test) from the exchange and set it to SKIP_INDEX as well.

cognifloyd:
I prefer explicit support in the deployment script versus an obscure, undocumented hack in the CircleCI config.

cognifloyd:
As far as `master` vs other branch names, we should make sure that pack maintainers can choose a different branch name: https://github.com/github/renaming

armab (changes requested):
Following https://github.com/StackStorm-Exchange/stackstorm-test/pull/16 discussion

I can see how flag to skip index deployment for testing purposes may be helpful. 

Instead of injecting it to CircleCI config, we may just add it to CircleCI UI to avoid possible mistakes in mass-overwriting `.circleci/config.yml` if that happens next time by a chance.

And so we'll also avoid read-only tokens https://github.com/StackStorm-Exchange/ci/blob/006d853df8bcd8b574a719095af09d11b5c843ca/tools/reset_github_user_token_and_update_circleci.sh#L91-L96

I however disagree that we should add support deploying non-`master` branch, just because we may need it sometime in the future (or may not need). This is counter-productive in both development and maintenance.

blag:
I agree somewhat with Eugen. We can pull the `SKIP_INDEX` from an environment variable, but we can set that within CircleCI's UI instead of customizing the CircleCI configs for these packs.

And this might be verging on bikeshedding, but I do think that we need a better name than `SKIP_INDEX` though. Something long like `DO_NOT_ADD_TO_EXCHANGE_INDEX` is a lot more descriptive over what the flag actually does. As it stands, a newbie may not understand that Exchange is just a slick UI against the `index` repository, and will have to dig a bit to figure out what that controls.

I do 100% agree that we should support deploying from a non-`master` branch name though. Not only because people are starting to switch to `main` instead of `master`, but also because it makes it more consistent if people install directly from a Git repo (`st2 pack install https://github.com/StackStorm-Exchange/test2`) versus installing from Exchange (`st2 pack install test2`). Right now, the first one would (I believe) install from the `aaa` branch, but the theoretical install from Exchange would install from the `master` branch. I'm not 100% sure about this, so please correct me if I'm wrong.

Also, I believe that the deploy step also collects some data about the pack when it generates the index (eg: number and name of actions, aliases, rules, policies, sensors, and triggers, as well as pack metadata from `pack.yaml`, the current pack version, and all available pack versions). So if the deploy step only collects data from `master`, then the content in the non-`master` default branch may not match what is advertised in Exchange.

cognifloyd:
Cool.. I like the idea of using a variable added via CircleCI ui.
I'll also rename the var as suggested.

armab:
We're talking about index generation/deployment script here and not the pack installation.

The only 2 repositories that have non-default `master` branch (`aaa` for testing purposes) are `test` and `test2`.
They're just for internal testing and we don't deploy these repositories to Exchange and don't plan to.

With that, non-master branch deployment would be just dead code if we merge it here.
I suggest to split this into a separated PR if we may need it sometime in the future.

armab:
From what I understand @cognifloyd initially opened this PR when realized that `test` and `test2` repos are failing to deploy to StackStorm Exchange index, until we discovered that these repositories should never present in that index.

So I'm not sure what we're trying to solve here.

blag:
These two packs were failing to deploy due to an indirect and undocumented bug/"feature" that we may not want to have anymore - the restriction that the deploy step only publishes the `master` branch.

Now, I'm not presently aware of any Exchange pack that publishes from a non-`master` branch, so it may not feel like removing that restriction is very useful to anybody. But, as @cognifloyd pointed out, there is a movement in the open source community to use words that are more specifically descriptive and less mired in historical injustices than words that have been used in the past. In the Django project, the terms master/slave when referring to databases were renamed to leader/follower, which is not only more accurate but also have less baggage than the words "master" and "follower". Similarly, in the spirit of inclusivity, there is a movement to rename the "main" branch where development happens away from `master` to `main`.

Since we are now aware that the default branch names for many projects including our own might be changing in the future, it seems like a good time to remove the artificial restriction on what branches can be deployed to Exchange. As part of this, instead of relying on indirect and undocumented behavior to seemingly accidentally not deploy to Exchange, it's a good time to improve the code to _explicitly_ skip deploying to Exchange, so future developers can instantly understand why these packs are not deploying to Exchange.

Just because we do not currently have any packs deploying from a non-`master` branch does not mean that it was always intended for that to be the case. For all we/I know, that could have been an oversight of previous maintainers. Now that it's explicit behavior it's easier to understand, reason about, and correct any future bugs in our deployment code.

cognifloyd:
I opened this PR for several reasons. `test` packs are only part (albeit a large part) of those reasons.

1. I made quite a few changes to exchange CI recently, including (a) drop python2 tests, (b) prevent PAT tokens from expiring, (c) standardize CircleCI config across all packs as much as is feasible, etc.
2. The `test` and `test2` packs would be ideal for testing changes like these to the CI infrastructure. That is the purpose listed in the README (see [1] below) for each of those packs. I could not use those packs, however, because ALL of the CircleCI config was out-of-date. They didn't even have CircleCI 2.0 config, and some of the changes were mistakenly added to the master branch. So, I had to use several other packs that were live on the exchange to test changes to the CI infrastructure which carried the risk of potentially breaking those packs if someone happened to install it during one of my tests.
3. The `deploy` workflow does more than just update the index. If we arbitrarily block the `deploy` workflow just to avoid updating the index, then I still have to use live packs to test changes to that `deploy` workflow. So, the `test` and `test2` packs must run the `deploy` workflow as well to meet their purpose (see [1] below) of testing the CI infrastructure.
4. For the `test` packs, exclusion from the index was a quirky accident based on obscure misuse of the CI config. This PR makes it possible to explicitly exclude those packs from the index. And it does so in a way that should be clear and easy-to-understand for future contributors.
5. Resolve #109 by issuing a good error message as early as possible if a PAT has expired.
6. The `test` packs also explicitly state (see [1] below) that their purpose is to ensure non-`master` branches work with our CI infrastructure. So, this PR is a *bugfix* PR, not an enhancement.
7. The `master` branch limitation is arbitrary and unnecessary. As @blag described, many projects are moving from `master` to `main` for very good reasons (we don't want to chase off existing and potential contributors). This bugfix PR removes the arbitrary limitation around branch naming so that we can use other branches at any point.
8. (9, 10, 11, ...) Please re-read what @blag wrote. Using descriptive names instead of `master` is extremely important across the open source world and beyond. I used reasons 1-4 to cover other reasons--mostly technical--for this change, but caring about people is even more important! So my reason number 7 is far more important than any of the other reasons. Without people we have no project--no matter how technically awesome the project is.

[1] Here is the relevant section from the README of both test packs (`test` and `test2`):

> NOTE: This pack is used for testing our pack Continous Integration system and uses a non-master default branch on Github (`aaa`) to verify that CI works with non-master default branches - please don't change.

.circle/deployment
```diff
@@ -21,6 +21,20 @@ fi
...
+
+DEFAULT_BRANCH=$(_gh_default_branch)
+# Give a good error message after the first use of github API in this script.
+if $? != 0; then
+    echo "Error retrieving data with github graphql API."
+    echo "The GitHub PAT might need to be regenerated:"
+    echo "https://github.com/settings/tokens/new?scopes=public_repo&description=CircleCI%3A%20stackstorm-${PACK_NAME}"
+else
+    echo "GitHub PAT is active."
+fi
```
cognifloyd, line 36: It is also possible to use `git` to query for the default branch: https://stackoverflow.com/a/44750379/1134951

But if we use `git` to retrieve the default branch, we still need an API call at this point in the script to validate that the PAT is still active.

Each API call counts towards our API usage limit. So, I would prefer that we use the return value of the API call for something productive; getting the default branch is productive.

blag:
I have added the `DO_NOT_ADD_TO_EXCHANGE_INDEX` environment variable to the CircleCI project settings for stackstorm-test, stackstorm-test2, and stackstorm-python3_test. It should have no effect on anything until this is merged.

.circle/deployment
```diff
@@ -21,6 +21,20 @@ fi
 # TODO: remove the install_gh call once circleci config is updated in all packs
 type gh &>/dev/null || ~/ci/.circle/install_gh
 
+# GH_TOKEN is used by gh which is used in functions.sh
+export GH_TOKEN=${MACHINE_PASSWORD}
+source ~/ci/tools/functions.sh
+
+DEFAULT_BRANCH=$(_gh_default_branch)
+# Give a good error message after the first use of github API in this script.
+if $? != 0; then
```
blag, line 30: Does this need to be in a test function?

```suggestion
if [[ $? != 0 ]]; then
```

cognifloyd: Good catch. Thanks!

cognifloyd:
@armab is the reasoning behind this PR clearer now? If so, please approve and merge.

blag (approved):

PR #110, opened 2021-02-15
Fixing exchange bootstrap script with new GitHub and CircleCI api changes

nmaludy:
GitHub and CircleCI broke their APIs.

GitHub doesn't allow us to create PATs anymore via the API, so we need to do that manually. I added a new environment variable `GITHUB_PACK_APT` that is checked and a helpful error message is output.

Also, for whatever reason the CircleCI API for following repos keeps returning 400 Bad Request. I've created a support ticket with them, so we'll see where that goes. It doesn't appear to affect anything as the project shows as "followed" by the stackstorm-neptr user.

On CircleCI i switched over to using their `Circle-Token` header instead of deprecated (no longer working) URL query parameter.

### DIFF
--- a/utils/exchange-bootstrap.sh
+++ b/utils/exchange-bootstrap.sh
@@ -38,6 +38,21 @@ REPO_DIR="/tmp/${REPO_NAME}"
 REPO_URL="https://${USERNAME}:${PASSWORD}@github.com/${EXCHANGE_ORG}/${REPO_NAME}"
 ALIAS_URL="https://${USERNAME}:${PASSWORD}@github.com/${EXCHANGE_ORG}/${REPO_ALIAS}"
 
+if [ -z "${GITHUB_PACK_PAT}" ]; then
+  echo "GitHub disabled the API allowing us to generate Personal Access Tokens for users"
+  echo "Please perform the following steps:"
+  echo "  1) login to GitHub as the 'stackstorm-neptr' account"
+  echo "  2) visit: https://github.com/settings/tokens"
+  echo "  3) click 'Genearte new token'"
+  echo "    3a) Note: CircleCI: ${REPO_NAME}"
+  echo "    3b) Select scopes: public_repo"
+  echo "    3c) Generate token"
+  echo "  4) Copy the new token and export it as the GITHUB_PACK_PAT shell variable"
+  echo "     export GITHUB_PACK_PAT='b4d3xxx'"
+  echo "  5) Re-run this script"
+fi
+
+
 # Check if the repo exists
 
 if git ls-remote "${REPO_URL}" > /dev/null 2>&1;
@@ -61,7 +76,7 @@ then
 	echo "The alias already exists, skipping the creation."
 else
 	echo "Creating an alias ${REPO_ALIAS} for ${REPO_NAME}."
-	curl -sS --fail -u "${USERNAME}:${PASSWORD}" -X POST --header "Content-Type: application/json" \
+	curl -v -sS --fail -u "${USERNAME}:${PASSWORD}" -X POST --header "Content-Type: application/json" \
 	-d '{"name": "'"${REPO_ALIAS}"'"}' \
 	"https://api.github.com/orgs/${EXCHANGE_ORG}/repos"
 fi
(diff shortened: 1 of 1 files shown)


utils/exchange-bootstrap.sh
```diff
@@ -119,42 +135,53 @@ then
...
+# echo "CircleCI: Following the project"
+# curl -v -sS --fail -X POST \
+#      --header "Circle-Token: ${CIRCLECI_TOKEN}" \
+#      "https://circleci.com/api/v1.1/project/github/${EXCHANGE_ORG}/${REPO_NAME}/follow"
+
 
 # CircleCI: upload the read-write key
 echo "CircleCI: Adding read-write SSH key"
 curl -sS --fail -X POST --header "Content-Type: application/json" \
+  --header "Circle-Token: ${CIRCLECI_TOKEN}" \
```
cognifloyd, line 149: inconsistent whitespace (tabs vs spaces strikes again)

utils/exchange-bootstrap.sh
```diff
@@ -119,42 +135,53 @@ then
...
 echo "CircleCI: Setting credentials (machine login and user-scoped token)"
 curl -sS --fail -X POST --header "Content-Type: application/json" \
+  --header "Circle-Token: ${CIRCLECI_TOKEN}" \
 	-d '{"name":"MACHINE_USER", "value":"'"${USERNAME}"'"}' \
-	"https://circleci.com/api/v1.1/project/github/${EXCHANGE_ORG}/${REPO_NAME}/envvar?circle-token=${CIRCLECI_TOKEN}"
+	"https://circleci.com/api/v1.1/project/github/${EXCHANGE_ORG}/${REPO_NAME}/envvar"
 curl -sS --fail -X POST --header "Content-Type: application/json" \
-	-d '{"name":"MACHINE_PASSWORD", "value":'"$(cat "/tmp/${PACK}_user_token")"'}' \
-	"https://circleci.com/api/v1.1/project/github/${EXCHANGE_ORG}/${REPO_NAME}/envvar?circle-token=${CIRCLECI_TOKEN}"
+  --header "Circle-Token: ${CIRCLECI_TOKEN}" \
```
cognifloyd, line 160: `s/^ */\t/`

utils/exchange-bootstrap.sh
```diff
@@ -119,42 +135,53 @@ then
...
+	-d '{"name":"MACHINE_PASSWORD", "value":"'"$(cat "/tmp/${PACK}_user_token")"'"}' \
+	"https://circleci.com/api/v1.1/project/github/${EXCHANGE_ORG}/${REPO_NAME}/envvar"
 
 # CircleCI: Enable builds for pull requests from forks
 echo "CircleCI: Enabling builds for pull requests from forks"
 curl -sS --fail -X PUT --header "Content-Type: application/json" \
+  --header "Circle-Token: ${CIRCLECI_TOKEN}" \
```
cognifloyd, line 167: `s/^ */\t/`

utils/exchange-bootstrap.sh
```diff
@@ -38,6 +41,23 @@ REPO_DIR="/tmp/${REPO_NAME}"
 REPO_URL="https://${USERNAME}:${PASSWORD}@github.com/${EXCHANGE_ORG}/${REPO_NAME}"
 ALIAS_URL="https://${USERNAME}:${PASSWORD}@github.com/${EXCHANGE_ORG}/${REPO_ALIAS}"
 
+# Generating GitHub PAT (personal access tokens) can no longer be done via the API
+# broken: https://developer.github.com/changes/2/#--deprecating-oauth-authorization-api
```
cognifloyd, line 45: We can probably reuse `tools/reset_github_user_token_and_update_circleci.sh` here once #107 is merged.
Something like this:

`tools/reset_github_user_token_and_update_circleci.sh --set-user ${PACK}`

utils/exchange-bootstrap.sh
```diff
@@ -119,42 +135,53 @@ then
...
 curl -sS --fail -X POST --header "Content-Type: application/json" \
+  --header "Circle-Token: ${CIRCLECI_TOKEN}" \
 	-d '{"name":"MACHINE_USER", "value":"'"${USERNAME}"'"}' \
-	"https://circleci.com/api/v1.1/project/github/${EXCHANGE_ORG}/${REPO_NAME}/envvar?circle-token=${CIRCLECI_TOKEN}"
+	"https://circleci.com/api/v1.1/project/github/${EXCHANGE_ORG}/${REPO_NAME}/envvar"
 curl -sS --fail -X POST --header "Content-Type: application/json" \
-	-d '{"name":"MACHINE_PASSWORD", "value":'"$(cat "/tmp/${PACK}_user_token")"'}' \
-	"https://circleci.com/api/v1.1/project/github/${EXCHANGE_ORG}/${REPO_NAME}/envvar?circle-token=${CIRCLECI_TOKEN}"
+  --header "Circle-Token: ${CIRCLECI_TOKEN}" \
+	-d '{"name":"MACHINE_PASSWORD", "value":"'"$(cat "/tmp/${PACK}_user_token")"'"}' \
```
cognifloyd, line 161:
```suggestion
	-d '{"name":"MACHINE_PASSWORD", "value":"'"${GITHUB_PACK_PAT}"'"}' \
```

utils/exchange-bootstrap.sh
```diff
@@ -79,21 +99,17 @@ curl -sS --fail -u "${USERNAME}:${PASSWORD}" -X POST --header "Content-Type: app
 	"https://api.github.com/repos/${EXCHANGE_ORG}/${REPO_NAME}/keys"
 
 # GitHub: create a user-scope token
-echo "Github: Creating a Github user-scoped token"
-curl -sS --fail -u "${USERNAME}:${PASSWORD}" -X POST --header "Content-Type: application/json" \
-	-d '{"scopes": ["public_repo"], "note": "CircleCI: '"${REPO_NAME}"'"}' \
-	"https://api.github.com/authorizations" | jq ".token" > "/tmp/${PACK}_user_token"
-
+echo -n "${GITHUB_PACK_PAT}" > "/tmp/${PACK}_user_token"
```
cognifloyd, line 102: Drop writing this to a file. When it's already in a var just use it from the var.

utils/exchange-bootstrap.sh
```diff
@@ -119,42 +135,53 @@ then
...
 
 # CircleCI: upload the read-write key
 echo "CircleCI: Adding read-write SSH key"
 curl -sS --fail -X POST --header "Content-Type: application/json" \
+  --header "Circle-Token: ${CIRCLECI_TOKEN}" \
 	-d '{"hostname":"github.com","private_key":"'"$(cat "/tmp/${PACK}_rsa")"'"}' \
-	"https://circleci.com/api/v1.1/project/github/${EXCHANGE_ORG}/${REPO_NAME}/ssh-key?circle-token=${CIRCLECI_TOKEN}"
+	"https://circleci.com/api/v1.1/project/github/${EXCHANGE_ORG}/${REPO_NAME}/ssh-key"
 
 # CircleCI: specify the credentials (the machine login and the new user-scope token)
```
cognifloyd, line 153: Setting MACHINE_USER and MACHINE_PASSWORD is handled well by `tools/reset_github_user_token_and_update_circleci.sh`. We should standardize on one script so we don't duplicate these calls in both.

cognifloyd (commented):
Here are a variety of thoughts. Mostly I think we need to use `tools/reset_github_user_token_and_update_circleci.sh` or extract the common logic into a separate utility shared by both tools.

utils/exchange-bootstrap.sh
armab: We're using 2 whitespaces across the st2 fleet for bash/script files and it indeed would be great following that in all the other places.

Mixing both (like this script) is indeed confusing.

cognifloyd (commented):
A couple of dangling references to `GITHUB_PACK_PAT` need to be removed.

cognifloyd (approved):

PR #113, opened 2021-02-18
Fix deployment if git checkout becomes unclean

cognifloyd:
The basic logic of the `deployment` script is:

1. get the list of commit hashes that modified `pack.yaml`
2. loop through each of those commit hashes
   - check out `pack.yaml` from that commit
   - if the version tag listed in `pack.yaml` is missing, tag the commit and push the tag
3. checkout `master` branch again (or whichever branch is default)
4. prepare/push the exchange index update

The checkout can become dirty in step 2 (see below).
This PR fixes step 3, by adding `--force` to the `git checkout` so it doesn't die if the checkout is dirty.

The checkout might be unclean if a script fails to parse old pack.yaml files.

This happens, for example, with the test pack. The test pack moved from version `0.3` to version `0.3.0` when semantic versioning was introduced/enforced. But `0.3` is not a valid semantic version, so one of the scripts ends up collapsing with the old version number.

The script actually handles this case correctly and just continue past the bad tag, but the git checkout is no longer clean so checking out `master` fails, which stops the script earlier than we intended.

### DIFF
--- a/.circle/deployment
+++ b/.circle/deployment
@@ -70,7 +70,7 @@ done
 
 # Afer pushing the version, make sure we check out the latest commit
 # This is important so index reflects latest changes
-git checkout ${DEFAULT_BRANCH}
+git checkout --force ${DEFAULT_BRANCH}
 git reset --hard ${DEFAULT_BRANCH}
 
 # Verify there are no unstaged changes


cognifloyd:
The key error this is resolving is:
```
+ git checkout -q ecda87378ea67f738205e2783843b2ee06ca9699 pack.yaml
++ /home/circleci/virtualenv/bin/python /home/circleci/ci/.circle/semver.py pack.yaml
+ VERSION=0.3.0
+ [[ -n 0.3.0 ]]
++ git rev-parse -q --verify refs/tags/v0.3.0
+ [[ -z b278915eccefdb01972673bad9903d58fdb7b91d ]]
+ [[ -n 0.3.0 ]]
++ git ls-remote origin refs/tags/v0.3.0
+ [[ -z b278915eccefdb01972673bad9903d58fdb7b91d	refs/tags/v0.3.0 ]]
+ git checkout master
error: Your local changes to the following files would be overwritten by checkout:
	pack.yaml
Please commit your changes or stash them before you switch branches.
Aborting
Command "/home/circleci/ci/.circle/deployment" failed with non-zero exit code, retrying...
```
https://app.circleci.com/pipelines/github/StackStorm-Exchange/stackstorm-test/10/workflows/eb10bc98-e863-4094-8e4f-b6be56734589/jobs/170/parallel-runs/0/steps/0-105

On subsequent retries, everything blows up because the git checkout is not clean.

cognifloyd:
For context, the basic logic of the `deployment` script is:

1. get the list of commit hashes that modified `pack.yaml`
2. loop through each of those commit hashes
   - check out `pack.yaml` from that commit
   - if the version tag listed in `pack.yaml` is missing, tag the commit and push the tag
3. checkout `master` branch again (or whichever branch is default)
4. prepare/push the exchange index update

The checkout becomes dirty in step 2.
This PR fixes step 3, by adding `--force` to the `git checkout` so it doesn't die.

cognifloyd:
We need to merge this, and then re-run this job: https://app.circleci.com/pipelines/github/StackStorm-Exchange/stackstorm-test/10/workflows/eb10bc98-e863-4094-8e4f-b6be56734589/jobs/170

blag:
What happens if the commit that changes `pack.yaml` is not the last commit in the PR?

That's a somewhat rare but still unsolved problem in our pack deployment.

Not saying we need to solve that with this PR, but I want to make sure that this PR doesn't make that situation worse. And I haven't reviewed this yet (sorry on mobile at the moment) so if I'm off base here then please ignore.

cognifloyd:
This script is triggered when pushing to the master/default branch.
So, it should be the latest commit. We do not do a `git fetch` anywhere, so `master`/default branch should point to the same commit for the duration of the job, even if other PRs are merged around the same time.

If multiple commits are merged at the same time, then I suppose there could still be a race condition, but I'm not concerned about it at this point.
If we have commits A and B (A is parent of B), and the deploy job for A starts cloning the repo after B is present, then we might have two deploy jobs that each run based on commit B. That shouldn't be a problem because the tags list should be the same either way.

If we want to tackle that possible race condition, that would be a different PR.

PR #117, opened 2021-11-22
Updates to ci/gha

lm-ydubler:
This PR includes updates to various files in the gha branch in order to convert the incubator to GHA & get it up and running.

### DIFF
--- a/.circle/Makefile
+++ b/.circle/Makefile
@@ -1,5 +1,5 @@
 ROOT_DIR ?= $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
-NUM_CHANGED_FILES := $(shell $(env.ACTION_PATH)/../../utils/git-changes files | wc -w)
+NUM_CHANGED_FILES := $(shell $(CI_DIR)/utils/git-changes files | wc -w)
 NUM_CHANGED_PY := $(shell $(CI_DIR)/utils/git-changes py | wc -w)
 NUM_CHANGED_YAML := $(shell $(CI_DIR)/utils/git-changes yaml | wc -w)
 NUM_CHANGED_JSON := $(shell $(CI_DIR)/utils/git-changes json | wc -w)
--- a/.circle/Makefile
+++ b/.circle/Makefile
@@ -1,9 +1,13 @@
 ROOT_DIR ?= $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
-NUM_CHANGED_FILES := $(shell /home/runner/work/_actions/lm-ydubler/ci/gha/utils/git-changes files | wc -w)
-NUM_CHANGED_PY := $(shell $(CI_DIR)/utils/git-changes py | wc -w)
-NUM_CHANGED_YAML := $(shell $(CI_DIR)/utils/git-changes yaml | wc -w)
-NUM_CHANGED_JSON := $(shell $(CI_DIR)/utils/git-changes json | wc -w)
-CHANGED_DIRECTORIES := $(shell $(CI_DIR)/utils/git-changes directories)
+#NUM_CHANGED_PY := $(shell $(CI_DIR)/utils/git-changes py | wc -w)
+#NUM_CHANGED_YAML := $(shell $(CI_DIR)/utils/git-changes yaml | wc -w)
+#NUM_CHANGED_JSON := $(shell $(CI_DIR)/utils/git-changes json | wc -w)
+#CHANGED_DIRECTORIES := $(shell $(CI_DIR)/utils/git-changes directories)
+NUM_CHANGED_FILES := $(shell /home/runner/work/_actions/lm-ydubler/ci/gha/utils/git-changes-files | wc -w)
+NUM_CHANGED_PY := $(shell /home/runner/work/_actions/lm-ydubler/ci/gha/utils/git-changes-py | wc -w)
+NUM_CHANGED_YAML := $(shell /home/runner/work/_actions/lm-ydubler/ci/gha/utils/git-changes-yaml | wc -w)
+NUM_CHANGED_JSON := $(shell /home/runner/work/_actions/lm-ydubler/ci/gha/utils/git-changes-json | wc -w)
+CHANGED_DIRECTORIES := $(shell /home/runner/work/_actions/lm-ydubler/ci/gha/utils/git-changes-directories)
 VIRTUALENV_DIR ?= virtualenv
 ST2_REPO_PATH ?= /tmp/st2
 ST2_REPO_BRANCH ?= master
@@ -294,7 +298,7 @@ requirements-ci:
 	@echo
 	. $(VIRTUALENV_DIR)/bin/activate && $(VIRTUALENV_DIR)/bin/pip install --upgrade "pip==20.3.3"
 	. $(VIRTUALENV_DIR)/bin/activate && $(VIRTUALENV_DIR)/bin/pip install --cache-dir $(HOME)/.pip-cache -q -r /home/runner/work/_actions/lm-ydubler/ci/gha/.circle/requirements-dev.txt
-	. $(VIRTUALENV_DIR)/bin/activate && $(VIRTUALENV_DIR)/bin/pip install --cache-dir $(HOME)/.pip-cache -q -r $(CI_DIR)/.circle/requirements-pack-tests.txt
+	. $(VIRTUALENV_DIR)/bin/activate && $(VIRTUALENV_DIR)/bin/pip install --cache-dir $(HOME)/.pip-cache -q -r /home/runner/work/_actions/lm-ydubler/ci/gha/.circle/.circle/requirements-pack-tests.txt
 
 .PHONY: virtualenv
 virtualenv: $(VIRTUALENV_DIR)/bin/activate
--- a/.circle/Makefile
+++ b/.circle/Makefile
@@ -1,5 +1,5 @@
 ROOT_DIR ?= $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
-NUM_CHANGED_FILES := $(shell /home/runner/work/_actions/lm-ydubler/ci/gha/utils/git-changes files | wc -w)
+NUM_CHANGED_FILES := $(shell $(CI_DIR)/utils/git-changes files | wc -w)
 NUM_CHANGED_PY := $(shell $(CI_DIR)/utils/git-changes py | wc -w)
 NUM_CHANGED_YAML := $(shell $(CI_DIR)/utils/git-changes yaml | wc -w)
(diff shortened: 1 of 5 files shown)


gha/apt-dependencies/action.yaml
```diff
@@ -29,22 +30,24 @@ runs:
...
       env:
         CACHE_HIT: ${{steps.cache-apt-deps.outputs.cache-hit}}
-        APT_PACKAGES_FILE_PATH: ${{github.action_path + '/apt-packages.txt'}}
+        APT_PACKAGES_FILE_PATH:  ${{ format('{0}{1}', github.action_path, '/apt-packages.txt') }}
       run: |
         if [[ "$CACHE_HIT" != 'true' ]]; then
         sudo apt-get -y update
         fi
-        APT_PACKAGES=$(grep -v '^#' "${APT_PACKAGES_FILE_PATH}" | xargs echo -n)
+        APT_PACKAGES=$(grep -v '^#' "${{env.APT_PACKAGES_FILE_PATH}}" | xargs echo -n)
```
cognifloyd, line 46: The `${{ }}` syntax gets substituted by github's workflow engine. I used `${ }` because I expected bash to handle substituting the var. Isn't `$APT_PACKAGES_FILE_PATH` available here?

If using `env:` to add vars is not working, then several things in the Makefile will be broken in later steps. So, I think we need to figure that out before continuing to much farther into the Makefile.

gha/py-dependencies/action.yaml
```diff
@@ -7,10 +7,12 @@ description: |
   StackStorm-Exchange/ci/gha/apt-dependencies.
 author: StackStorm
 
-inputs:
+
+s:
```
cognifloyd, line 11: Typo?
```suggestion
inputs:
```

gha/py-dependencies/action.yaml
```diff
@@ -55,48 +59,52 @@ runs:
         ./scripts/github/install-virtualenv.sh
         PIP_VERSION=$(grep '^PIP_VERSION' Makefile | awk '{print $3}')
         echo "::set-output name=pip-version::${PIP_VERSION}"
-        ${VIRTUALENV_DIR}/bin/pip install -U "pip==${PIP_VERSION}" setuptools
+        ${{env.VIRTUALENV_DIR}}/bin/pip install -U "pip==${PIP_VERSION}" setuptools
```
cognifloyd, line 62: The `${{ }}` vs `${ }` bits are more problematic here. If VIRTUALENV_DIR is not available here, then install-virtualenv.sh might not work correctly.

.circle/Makefile
```diff
@@ -1,9 +1,13 @@
 ROOT_DIR ?= $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
-NUM_CHANGED_FILES := $(shell $(CI_DIR)/utils/git-changes files | wc -w)
-NUM_CHANGED_PY := $(shell $(CI_DIR)/utils/git-changes py | wc -w)
-NUM_CHANGED_YAML := $(shell $(CI_DIR)/utils/git-changes yaml | wc -w)
-NUM_CHANGED_JSON := $(shell $(CI_DIR)/utils/git-changes json | wc -w)
-CHANGED_DIRECTORIES := $(shell $(CI_DIR)/utils/git-changes directories)
+#NUM_CHANGED_PY := $(shell $(CI_DIR)/utils/git-changes py | wc -w)
```
cognifloyd, line 2: If CI_DIR is not populated correctly, that is concerning. Hmm.

gha/test/action.yaml
```diff
@@ -37,14 +38,15 @@ runs:
...
 
-    - name: 
+    - name:
+      shell: bash
       env:
         VIRTUALENV_DIR: "~/virtualenv"
         FORCE_CHECK_ALL_FILES: true if on default branch
         ROOT_DIR: pack directory
-        CI_DIR: ${{ github.action_path + "/../.." }}
+        CI_DIR: ${{ format('{0}{1}', github.action_path, '/../..') }}
```
cognifloyd, line 47: Here is where CI_DIR is populated - I kind of guessed on the relative path though, so feel free to adjust this if it's not right.

gha/apt-dependencies/action.yaml
cognifloyd: https://docs.github.com/en/actions/learn-github-actions/environment-variables

gha/py-dependencies/action.yaml
lm-ydubler: Yes, typo. Complete accident. Didn't even know it happened. Resolved in a recent commit.

gha/apt-dependencies/action.yaml
lm-ydubler: Ok, I'll scan the file differentiating between those two syntaxes and make sure they link up correctly and run some tests.

.circle/Makefile
lm-ydubler: It might have been populating correctly but possibly just had the wrong value. GHA doesn't allow string concatenation using a + so I improvised to use the format() function. So I'll look it over and make sure it all matches up knowing what the correct directory should be as a fail safe.

gha/checkout/action.yaml
```diff
@@ -23,15 +23,17 @@ runs:
...
     - name: Configure git user
+      shell: bash
       env:
         GIT_USER_NAME: "${{ inputs.git_user_name }}"
         GIT_USER_EMAIL: "${{ inputs.git_user_email }}"
       run: |
-        git config --global user.name "${GIT_USER_NAME}"
-        git config --global user.email "${GIT_USER_EMAIL}"
+        git config --global user.name "${{env.GIT_USER_NAME}}"
+        git config --global user.email "${{env.GIT_USER_EMAIL}}"
```
cognifloyd, line 32: We can use `env` to check out the env and make sure our env vars are available.
```suggestion
        env
        git config --global user.name "${GIT_USER_NAME}"
        git config --global user.email "${GIT_USER_EMAIL}"
```

gha/checkout/action.yaml
```diff
@@ -23,15 +23,17 @@ runs:
...
 
     - name: Checkout pack repo
       uses: actions/checkout@v2
+      with:
         path: pack
         depth: 1
```
cognifloyd, line 38: Oops. I used the wrong param here.
```suggestion
        fetch-depth: 1
```
same with the other checkout lines below.

gha/apt-dependencies/action.yaml
lm-ydubler: Ok, so here is a test output for using `${ }` vs `${{ }}` :
single-vs-double

The single brackets use the pure string whereas the double brackets appear to format it as the actual path. So, this is good to know.

gha/test/action.yaml
lm-ydubler: single-vs-double

gha/py-dependencies/action.yaml
lm-ydubler: Just so you see the comparison here too:

.circle/Makefile
lm-ydubler: And just so you know, this is a testing PR and has a million commits a second. When it checks out , I'll create the final-draft PR which will have a handful of commits just so that it comes across clean.

gha/apt-dependencies/action.yaml
cognifloyd: When you use the var, bash will expand `~` to `/home/runner`, so I thought they would be effectively the same... apparently not. We need the `env:` block to work..., so I have another suggestion...

.circle/Makefile
cognifloyd: Yes. I expected that. But having it in a PR allows for some discussion about the issues you are working on 😄

gha/py-dependencies/action.yaml
```diff
@@ -55,27 +59,31 @@ runs:
...
 
     - name: Install StackStorm requirements
       # this should run in the st2 checkout
+      shell: bash
       working-directory: st2
       env:
         VIRTUALENV_DIR: "~/virtualenv"
```
cognifloyd, line 69: OK. Since tilde expansion isn't happening like I thought, then maybe we can do this instead:
```suggestion
        VIRTUALENV_DIR: ${{ env.HOME }}/virtualenv
```

To do that, we need to add a step at the beginning of the job like this:
```yaml
- name: add HOME to env context
  shell: bash
  run: echo ::set-env name=HOME::$HOME
```

gha/py-dependencies/action.yaml
```diff
@@ -55,27 +59,31 @@ runs:
...
+      shell: bash
       env:
         VIRTUALENV_DIR: "~/virtualenv"
         ACTION_PATH: ${{ github.action_path }}
         ST2_REPO_PATH: st2
+        CI_DIR: ${{ format('{0}{1}', github.action_path, '/../..') }}
       # TODO: This Makefile has CircleCI assumptions
       run: |
-        cp ${ACTION_PATH}/../../.circle/Makefile .
+        cp ${{env.ACTION_PATH}}/../../.circle/Makefile .
```
cognifloyd, line 82: It turns out, github already exposes action_path as an env var, so we can remove ACTION_PATH from the env block and just use `$GITHUB_ACTION_PATH`
```suggestion
        cp ${GITHUB_ACTION_PATH}/../../.circle/Makefile .
```

.circle/Makefile
cognifloyd: I'm going to consolidate some of your commits and add them to the gha branch of StackStorm-Exchange/ci.

gha/py-dependencies/action.yaml
lm-ydubler: Just for reference:
single-vs-double

cognifloyd:
@lm-ydubler I just consolidated a bunch of the changes into the gha branch. Could you merge that into your gha branch to reduce the diff?

gha/py-dependencies/action.yaml
cognifloyd: I just pushed this into change to the upstream gha branch.

gha/py-dependencies/action.yaml
cognifloyd: I just pushed this to the upstream gha branch

armab:
Thanks, folks for the work you're doing :+1: 

To connect the dots, just left some notes on the [ Proposal: Use Github Actions for StackStorm-Exchange CI and Maintenance #63](https://github.com/StackStorm/discussions/issues/63) which is relevant to this PR.

cognifloyd:
OK. The build_and_test workflow is good to go on the gha branch. I'm going to close this PR now. We'll open up another PR for any collaboration on the deploy workflow, and then one final PR to merge the gha branch into master.

PR #120, opened 2021-12-07
Updates to Tag Release Workflow

lm-ydubler:


cognifloyd (commented):
Let's separate this into multiple steps so that we can see the progression in the gorgon actions UI more easily. We don't have to loop over anything so this will be more straight forward than the update-index workflow was.

Looking around, i think we can simplify extracting the version by using `yq`. And maybe we can put the conditional on the workflow itself instead of in the bash script.

wdyt?

PR #126, opened 2021-12-20
Handle py2-only packs more gracefully in GHA

cognifloyd:
Basically, this converts the `.circle/exit_on_py3_checks` script to GHA.

This will allow these py2-only packs to update their metadata in the index as it worked under CircleCI:
https://github.com/StackStorm-Exchange/stackstorm-clicrud
https://github.com/StackStorm-Exchange/stackstorm-network_essentials
https://github.com/StackStorm-Exchange/stackstorm-vdx
https://github.com/StackStorm-Exchange/stackstorm-vdx_vtep

To see how this worked on CircleCI, check out: https://app.circleci.com/pipelines/github/StackStorm-Exchange/stackstorm-clicrud/51/workflows/a2a4082c-0f96-4ce8-916f-9ea3522956e4/jobs/311
```
/home/circleci/repo/actions/ops_command.py:56:28: F821 undefined name 'unicode'
make: *** [Makefile:78: .flake8] Error 1
Original script exit code: 2
Ignoring failures since pack doesn't declare Python 3 support in pack.yaml
```

I also exposed a way for pack authors to pin the st2 version used for tests instead.

PR #127, opened 2021-12-20
Handle py2-only packs more gracefully in GHA

cognifloyd:
oops. merged #126 to wrong branch.


cognifloyd:
ugh. This won't work because composite actions don't support continue-on-error :/

winem (approved):

PR #128, opened 2021-12-20
Add 'pip list' to show versions step

lm-ydubler:
Added pip list


cognifloyd:
Thanks! Merged as part of #127

PR #130, opened 2021-12-23
GHA: Cleanup/update requirements

cognifloyd:
The requirements files were out-of-date and conflicted with the requirements pinned in st2.git.
Instead of pinning them here, let's just pull the pinned version from our st2 clone.

Also, I deleted a couple of requirements files that aren't even used by the gha workflow (ie I shouldn't have copied them from the `.circle` folder).

Thanks @lm-ydubler for helping me research where these requirements differ.


winem:
Very nice again. I know it's not introduced by you but I saw a typo on the top comment in .github/actions/py-dependencies/requirements-dev.txt, check is missing a c. Would be a good chance to fix that, too.

.github/actions/py-dependencies/requirements-dev.txt
```diff
@@ -1,9 +1,8 @@
 # Needed for chek and lint scripts
```
winem, line 1: Not introduced by this PR but it's a good chance to fix the typo (missing c) in check.

winem (approved):

PR #135, opened 2022-12-08
pin GHA runners to ubuntu-20.04

cognifloyd:
This mirrors StackStorm/st2#5832 and fixes the index builds that have been failing.

See: https://github.com/StackStorm-Exchange/index/actions


amanda11 (approved):
LGTM

nzlosh (commented):
LGTM

PR #137, opened 2022-12-15
Add "Enable Pack Build and Test" Workflow

### LINKED ISSUE #136: Echange Packs GH Action workflows disabled after 60days of inactivity
The problem when there were no commits in the repository for `60 days`, - Github Automatically disables CI for it.
It happens all the time when we review new contributions in Exchange and figure out after merging that CI didn't run resulting in post-merge linting errors.

Possible solutions:

1)  https://github.com/marketplace/actions/keepalive-workflow heart beat commits .. :nauseated_face: sounds like a hack
2) Per @nzlosh  There's a REST API endpoint that can be called to enable workflows, so that'd be more robust.  https://docs.github.com/en/rest/actions/workflows?apiVersion=2022-11-28#enable-a-workflow  We could schedule an St2 workflow that iterates all exchange repos and calls the API to enable CI workflows. From the possible workarounds, this looks like the cleanest and most maintainable way to workaround the github limitation.

3) Considering the StackStorm instance is automatically started/stopped, 3rd option could be an AWS Lambda function trying to re-enable GH Action Workflows for the Exchange repos. Example solution https://stackoverflow.com/a/68591566/4533625

mamercad:
Add a Workflow which ensure that the "Build and Test" Workflows are enabled in all of StackStorm-Exchange; GitHub will disable all Workflows in a repository if [60d have gone by without activity](https://docs.github.com/en/actions/managing-workflow-runs/disabling-and-enabling-a-workflow); fixes #136.


.github/workflows/pack-enable_release-workflow.yaml
```diff
@@ -0,0 +1,145 @@
...
+            echo "${PACKS_ORG}/${repo_name}" >> ${GITHUB_WORKSPACE}/bad_pack_repos
+          elif [[ " ${EXCLUDE_PACKS} " =~ " ${repo_name#${PACKS_PREFIX}-} " ]]; then
+            rm -rf ${repo_name}
+            echo "DELETED clone of ${PACKS_ORG}/${repo_name}: pack excluded via index.git/v1/exclude_packs.txt"
+            echo "${PACKS_ORG}/${repo_name}" >> ${GITHUB_WORKSPACE}/bad_pack_repos
+          else
+            if [[ -f "./${repo_name}/.github/workflows/release.yaml" ]]; then
+              echo "GitHub release workflow manifest found for ${PACKS_ORG}/${repo_name}"
+
+              workflow_exists_http_code="$(curl \
```
cognifloyd, line 100: Is there a reason to use `curl` instead of `gh` (which is installed by default on GHA runners)? All of our other Github API calls use `gh`, so I would prefer consistency here.

cognifloyd (commented):
I understand this is a draft. Thanks for tackling this!

Here are a couple of thoughts.

Would you document (probably in a comment) what permissions are required to use this? What permissions should `${{ github.token }}` (or equivalent) have?

I suspect we will need to pass a token in as a secret. I, or one of the senior maintainers, can help setup that token so it can be accessed via GHA workflows.

.github/workflows/pack-enable_release-workflow.yaml
mamercad: > Is there a reason to use `curl` instead of `gh` (which is installed by default on GHA runners)? All of our other Github API calls use `gh`, so I would prefer consistency here.

Oh, good thought, `gh workflow enable` would be so much cleaner.

.github/workflows/enable-pack-release.yaml
```diff
@@ -0,0 +1,123 @@
...
+# Only allow one workflow run per branch at a time.
+# When one is already running, new runs will be pending.
+# https://github.blog/changelog/2021-04-19-github-actions-limit-workflow-run-or-job-concurrency/
+concurrency:
+  group: enable-pack-release-${{ github.ref }}
+  # Do not use cancel-in-progress! Or we might create a scenario where all workflows get canceled before they complete.
+
+permissions:
+  actions: write # Need to enable Actions
+  contents: read # Need to clone repositories
```
cognifloyd, line 25: These permissions apply to the auto-generated github.token. But, that token cannot get write access to any other repo. So, I think we'll need some other token for this, and probably stored as a secret in github. I can help create that - let me know what kind of token and what permissions are needed and I'll set it up.

.github/workflows/enable-pack-release.yaml
```diff
@@ -0,0 +1,123 @@
...
+            echo "${PACKS_ORG}/${repo_name}" >> ${GITHUB_WORKSPACE}/bad_pack_repos
+          else
+            #
+            #  This is not a bad or excluded Pack, ensure the Release workflow is active
+            #
+            echo "::group::Ensure Release workflow for ${PACKS_ORG}/${repo_name} is active"
+            if [[ -f "./${repo_name}/.github/workflows/release.yaml" ]]; then
+              echo "GitHub Release workflow manifest found for ${PACKS_ORG}/${repo_name}"
+                # NOTE: Unfortunately "gh workflow list" does not have --json
+                if ! gh workflow list ${PACKS_ORG}/${repo_name} --all | grep -E '^Release\b\s+active\b'; then
```
cognifloyd, line 108: If `gh workflow list` isn't giving good output, you could use `gh` to make the API calls more directly too.
Translating your old curl command would be something like: `gh api -X GET repos/{owner}/{repo}/actions/workflows/release.yaml --jq .state`

.github/workflows/pack-enable_release-workflow.yaml
```diff
@@ -0,0 +1,124 @@
...
+            echo "DELETED clone of ${PACKS_ORG}/${repo_name}: missing git tags"
+            echo "${PACKS_ORG}/${repo_name}" >> ${GITHUB_WORKSPACE}/bad_pack_repos
+          elif [[ ! -f "./${repo_name}/pack.yaml" ]]; then
+            rm -rf ${repo_name}
+            echo "DELETED clone of ${PACKS_ORG}/${repo_name}: missing pack.yaml"
+            echo "${PACKS_ORG}/${repo_name}" >> ${GITHUB_WORKSPACE}/bad_pack_repos
+          elif [[ " ${EXCLUDE_PACKS} " =~ " ${repo_name#${PACKS_PREFIX}-} " ]]; then
+            rm -rf ${repo_name}
+            echo "DELETED clone of ${PACKS_ORG}/${repo_name}: pack excluded via index.git/v1/exclude_packs.txt"
+            echo "${PACKS_ORG}/${repo_name}" >> ${GITHUB_WORKSPACE}/bad_pack_repos
```
cognifloyd, line 96: For the index CI, this allows excluding some test packs from the index. For this workflow, we still want those test packs to have CI working. So, I think you can drop this here and probably skip cloning the index repo.

.github/workflows/enable-pack-release.yaml
```diff
@@ -0,0 +1,123 @@
...
+            echo "DELETED clone of ${PACKS_ORG}/${repo_name}: missing git tags"
+            echo "${PACKS_ORG}/${repo_name}" >> ${GITHUB_WORKSPACE}/bad_pack_repos
+          elif [[ ! -f "./${repo_name}/pack.yaml" ]]; then
+            rm -rf ${repo_name}
+            echo "DELETED clone of ${PACKS_ORG}/${repo_name}: missing pack.yaml"
+            echo "${PACKS_ORG}/${repo_name}" >> ${GITHUB_WORKSPACE}/bad_pack_repos
+          elif [[ " ${EXCLUDE_PACKS} " =~ " ${repo_name#${PACKS_PREFIX}-} " ]]; then
+            rm -rf ${repo_name}
+            echo "DELETED clone of ${PACKS_ORG}/${repo_name}: pack excluded via index.git/v1/exclude_packs.txt"
+            echo "${PACKS_ORG}/${repo_name}" >> ${GITHUB_WORKSPACE}/bad_pack_repos
```
cognifloyd, line 99: For the index CI, this allows excluding some test packs from the index. For this workflow, we still want those test packs to have CI working. So, I think you can drop this here and probably skip cloning the index repo.

.github/workflows/enable-pack-release.yaml
mamercad: > If `gh workflow list` isn't giving good output, you could use `gh` to make the API calls more directly too. Translating your old curl command would be something like: `gh api -X GET repos/{owner}/{repo}/actions/workflows/release.yaml --jq .state`

Ooh, that's nice, thanks.

.github/workflows/enable-pack-release.yaml
mamercad: > These permissions apply to the auto-generated github.token. But, that token cannot get write access to any other repo. So, I think we'll need some other token for this, and probably stored as a secret in github. I can help create that - let me know what kind of token and what permissions are needed and I'll set it up.

Hrm, I guess fine-grained PATs are likely out which only leaves "Classic PATs":

```shell
❯ gh repo list
HTTP 401: Personal access tokens with fine grained access do not support the GraphQL API (https://api.github.com/graphql)
```

This seems to be the bare minimum:

image

```shell
❯ gh workflow list --repo mamercad/dotfiles
Lint Code Base  active  2404312
integration     active  32473778

❯ gh workflow disable 2404312 --repo mamercad/dotfiles
✓ Disabled Lint Code Base

❯ gh workflow enable 2404312 --repo mamercad/dotfiles
✓ Enabled Lint Code Base
```

mamercad: How's [this](https://github.com/StackStorm-Exchange/ci/pull/137/files#diff-fec4ffbf93caa5b3185269c68882ee80ed1d1bd95675ba488c9743c2f096c58cR15) look?

.github/workflows/enable-pack-release.yaml
```diff
@@ -0,0 +1,116 @@
...
+          elif [[ ! -f "./${repo_name}/pack.yaml" ]]; then
+            rm -rf ${repo_name}
+            echo "DELETED clone of ${PACKS_ORG}/${repo_name}: missing pack.yaml"
+            echo "${PACKS_ORG}/${repo_name}" >> ${GITHUB_WORKSPACE}/bad_pack_repos
+          else
+            #
+            #  This is not a bad Pack, ensure the Release workflow is active
+            #
+            echo "::group::Ensure Release workflow for ${PACKS_ORG}/${repo_name} is active"
+            RELEASE_WORKFLOW="repos/${PACKS_ORG}/${repo_name}/actions/workflows/release.yaml"
```
armab, line 95: Just a note, it's the `CI` (`build_and_test.yam`l) workflow that needs to be re-activated instead of "Release".

Screenshot 2022-12-17 at 14-53-05 Actions · StackStorm-Exchange_stackstorm-ansible

.github/workflows/enable-pack-build-and-test.yaml
```diff
@@ -0,0 +1,115 @@
...
+              #
+              ORG_REPO="${PACKS_ORG}/${repo_name}"
+              WORKFLOW="repos/${PACKS_ORG}/${repo_name}/actions/workflows/build_and_test.yaml"
+              WORKFLOW_NAME="CI/build_and_test"
+              echo "::group::Ensure ${WORKFLOW_NAME} workflow for ${PACKS_ORG}/${repo_name} is active"
+              if gh api --silent --method GET $WORKFLOW 2>/dev/null; then
+                echo "GitHub ${WORKFLOW_NAME} workflow found for ${PACKS_ORG}/${repo_name}"
+                if [[ "$(gh api --method GET $WORKFLOW --jq .state)" != "active" ]]; then
+                  echo "::notice::GitHub ${WORKFLOW_NAME} workflow for ${PACKS_ORG}/${repo_name} not active, enabling"
+                  gh workflow enable $WORKFLOW --repo ${PACKS_ORG}/${repo_name}
```
cognifloyd, line 99: I don't think the API path will work here. Should this be something like this?

```suggestion
                  gh workflow enable ${WORKFLOW_NAME} --repo ${PACKS_ORG}/${repo_name}
```

.github/workflows/enable-pack-build-and-test.yaml
```diff
@@ -0,0 +1,115 @@
...
+              echo "${PACKS_ORG}/${repo_name}" >> ${GITHUB_WORKSPACE}/bad_pack_repos
+            elif [[ ! -f "./${repo_name}/pack.yaml" ]]; then
+              rm -rf ${repo_name}
+              echo "DELETED clone of ${PACKS_ORG}/${repo_name}: missing pack.yaml"
+              echo "${PACKS_ORG}/${repo_name}" >> ${GITHUB_WORKSPACE}/bad_pack_repos
+            else
+              #
+              #  This is not a bad Pack, ensure the Build and Test workflow is active
+              #
+              ORG_REPO="${PACKS_ORG}/${repo_name}"
```
cognifloyd, line 91: It looks like this is unused.

cognifloyd (commented):
Looking good. Next time I have a moment I will make the PAT.

Does the user who makes the PAT need to be an admin? If not, I'll use the `stackstorm-neptr` bot account.

.github/workflows/enable-pack-build-and-test.yaml
```diff
@@ -0,0 +1,115 @@
+name: Enable Pack Build and Test
+
+on:
+  # First day of the month at midnight.
+  schedule:
+    - cron: "0 0 1 * *"
```
armab, line 6: If we'll run the re-enable workflow once a month, there's a ~29 days for packs CI to expire.
With that, I think a daily run would be sufficient to cover all the `60d` expiration cases.

.github/workflows/enable-pack-build-and-test.yaml
mamercad: > I don't think the API path will work here. Should this be something like this?

Ah yes, good catch.

armab:
From the https://github.com/StackStorm-Exchange/index/pull/27#issuecomment-1354073870

> I wonder if it would make more sense to schedule this workflow on https://github.com/StackStorm-Exchange/ci instead of on the index. The current workflow in this repo only has write access to the index repo, and nothing else. I think I'd like to keep it that way.

@cognifloyd @mamercad I think the corner case here is that this repository `ci` can expire the same way as packs, making this workflow non-running after the `60 days`.

With that, the difference between this repo and Index (https://github.com/StackStorm-Exchange/index/) is that Index repo CI won't expire due to [daily automated commits](https://github.com/StackStorm-Exchange/ci/commits/master).

.github/workflows/enable-pack-build-and-test.yaml
mamercad: It'll just be the top-level Workflow name I think:

```shell
❯ gh workflow list --repo stackstorm-exchange/stackstorm-jira
CI       active  16569399
Release  active  16569400
```

.github/workflows/enable-pack-build-and-test.yaml
mamercad: Good catch on my bikeshedding that didn't happen 😄

mamercad:
> From the [StackStorm-Exchange/index#27 (comment)](https://github.com/StackStorm-Exchange/index/pull/27#issuecomment-1354073870)
> 
> > I wonder if it would make more sense to schedule this workflow on https://github.com/StackStorm-Exchange/ci instead of on the index. The current workflow in this repo only has write access to the index repo, and nothing else. I think I'd like to keep it that way.
> 
> @cognifloyd @mamercad I think the corner case here is that this repository GH Workflows can expire the same way as packs, making this workflow non-running after the `60 days`.
> 
> With that, the difference between this repo and Index (https://github.com/StackStorm-Exchange/index/) is that Index repo CI won't expire due to [daily automated commits](https://github.com/StackStorm-Exchange/ci/commits/master).
> 
> 🤔

Ah yes, thinking... we found the "zero point" 😄

.github/workflows/enable-pack-build-and-test.yaml
```diff
@@ -0,0 +1,114 @@
...
+            else
+              #
+              #  This is not a bad Pack, ensure the Build and Test workflow is active
+              #
+              WORKFLOW_PATH="repos/${PACKS_ORG}/${repo_name}/actions/workflows/build_and_test.yaml"
+              WORKFLOW_NAME="CI"
+              echo "::group::Ensure ${WORKFLOW_NAME} workflow for ${PACKS_ORG}/${repo_name} is active"
+              if gh api --silent --method GET ${WORKFLOW_PATH} 2>/dev/null; then
+                echo "GitHub ${WORKFLOW_NAME} workflow found for ${PACKS_ORG}/${repo_name}"
+                if [[ "$(gh api --method GET ${WORKFLOW_PATH} --jq .state)" != "active" ]]; then
```
cognifloyd, line 96: What if we skipped this check and just blindly enable it every time this workflow runs? That way, the 60d clock would reset every time this runs.

I wonder if it could also self-enable? So, every time it runs it hits the API to reset the clock on the CI repo as well.

mamercad: Oh, that's interesting and might work, have you found docs around what things specifically constitute repository activity? I have not ... well, other than the obvious (commits ... and I'm with you, I don't like dummy commits).

mamercad:
> Looking good. Next time I have a moment I will make the PAT.
> 
> Does the user who makes the PAT need to be an admin? If not, I'll use the `stackstorm-neptr` bot account.

I don't think it does.

.github/workflows/enable-pack-build-and-test.yaml
mamercad: I wish we knew what types of things constituted "repository activity" ... I still haven't managed to find specific information on this. I'm hoping it's more than just "commits against the main branch" (because that'd mean we'd have to do dummy commits into it). What's the monitoring story like for StackStorm, maybe an alternative to dummy commits into `master` would be to simply monitor *this* job's enabled or disabled status with a "daily dead man's switch" or something along those lines.

mamercad:
> > Looking good. Next time I have a moment I will make the PAT.
> > Does the user who makes the PAT need to be an admin? If not, I'll use the `stackstorm-neptr` bot account.
> 
> I don't think it does.

Also, let me know if/when you're ready to merge this, I can rebase it beforehand.

.github/workflows/enable-pack-build-and-test.yaml
```diff
@@ -0,0 +1,116 @@
+name: Enable Pack Build and Test
+
+on:
+  # Daily at 3 in the morning
+  schedule:
+    - cron: "0 3 * * *"
```
cognifloyd, line 6: OK. I created the secret. I'm almost ready to merge this.

@armab I think once a month is enough, assuming that the API call resets the 60 day timer. That assumption is also baked into this:
https://stackoverflow.com/a/68591566/1134951
https://github.com/invenia/KeepActionsAlive

So, can we try once a month, and then if that turns out not to work, we can tweak this?

armab: > What if we skipped this check and just blindly enable it every time this workflow runs? That way, the 60d clock would reset every time this runs.

I hope it'll work, but I'd doubt it's possible to re-enable not yet disabled GH workflow, considering the solution mentioned there and this code: https://github.com/invenia/KeepActionsAlive/blob/main/src/app.py#L73

Otherwise, we'd see similar solutions in the wild when the github repo could re-enable itself, instead of using the https://github.com/marketplace/actions/keepalive-workflow self-commit approach.

More than that, the CI may expire after 2 months in this repo too. With that, I think there's a need to use this workflow in the https://github.com/stackStorm-exchange/index, as we have automated commits in there and confidence it won't expire.

But we can try as is in this repo and see how it goes to verify our guesses.

mamercad: > > What if we skipped this check and just blindly enable it every time this workflow runs? That way, the 60d clock would reset every time this runs.
> 
> I hope it'll work, but I'd doubt it's possible to re-enable not yet disabled GH workflow, considering the solution mentioned there and the code: https://github.com/invenia/KeepActionsAlive/blob/main/src/app.py#L73
> 
> Otherwise, we'd see similar solutions in the wild when the github repo could re-enable itself, instead of using the https://github.com/marketplace/actions/keepalive-workflow self-commit approach.
> 
> More than that, the GH Actions may expire after 2 months in this repo too. With that, I think there's a need to use this workflow in the https://github.com/stackStorm-exchange/index, as we have automated commits in there and confidence it won't expire.
> 
> But we can try as is in this repo and see how it goes to verify our guesses.

Yeah, my suspicion is that this "self-enabling" approach won't work either; using the index isn't a terrible approach since the index and ci repositories are functionally coupled anyhow, right?

cognifloyd: The index repo would be more convenient, true. But, I want to avoid adding a PAT on the index repo.

Right now, all the packs get pulled into the index CI. If someone compromises one of our packs, then the most they can do is mess up the index, they can't use the index to get to any other exchange packs.

If we add a PAT with write access to repos and workflows, then that widens such an attack vector. Then, if someone is able to compromise a pack and the index, they could jump from there to all the packs.

This CI repo is used in workflows across the exchange. Anyone that gets access to this repo already has access to all the packs. But, gaining access to any of the packs does not give any attack vector that could compromise this repo. So, I would prefer to have the PAT only on this repo.

(Hmm, maybe this workflow should avoid cloning all the packs, and rely on the GitHub API to look up everything about the available workflows).

If the API call doesn't work on this repo (I agree, there's a good chance it won't), then we can do one of three things:
- add dummy commits (doing it only in this repo, while ugly, is better than on all of the pack repos)
- add a workflow in st2 ci that deals with this repo (and only this repo, not the rest of the exchange)
- add a Circle I workflow that keeps this repo active.

I kind of like the idea of using CircleCI here. The token we add to CircleCI could use a fine grained token so that it can only hit the workflow enable endpoint on this one repo.

mamercad: > If we add a PAT with write access to repos and workflows

I don't think this would need write on the *repos* though, just on the workflows.

armab: All the good options :+1: 

- fine-grained PAT
- CircleCI
- automated commits in this repo

I'm fine either way

mamercad: Cool, deferring to @cognifloyd on which way to go!

cognifloyd: I'm merging it as is. Let's see how things go and then make changes in follow-ups

cognifloyd (approved):

cognifloyd:
@mamercad looks like there are some gotchas with the workflow. Could you take a look? I suspect we need to just change the working directory, or specify the repo on the command line:
https://github.com/StackStorm-Exchange/ci/actions/runs/3737834951/jobs/6343366911

mamercad:
> @mamercad looks like there are some gotchas with the workflow. Could you take a look? I suspect we need to just change the working directory, or specify the repo on the command line: https://github.com/StackStorm-Exchange/ci/actions/runs/3737834951/jobs/6343366911

Surely; the first thing that I've noticed is that it looks like [this](https://github.com/cli/cli/issues/5025) is a thing we're going to run into (I can reproduce it in my fork):

```shell
❯ gh workflow list --repo mamercad/stackstorm-exchange-ci      
Enable Pack Build and Test  active  43546553
CI - Build and Test         active  43398038
Test                        active  43324335

❯ gh workflow enable "Enable Pack Build and Test" --repo mamercad/stackstorm-exchange-ci
could not find any workflows named Enable Pack Build and Test

❯ gh workflow disable "Enable Pack Build and Test" --repo mamercad/stackstorm-exchange-ci
✓ Disabled Enable Pack Build and Test

❯ gh workflow enable "Enable Pack Build and Test" --repo mamercad/stackstorm-exchange-ci
✓ Enabled Enable Pack Build and Test
```

The second command there is problematic if the workflow is already enabled since it exits nonzero:

```
❯ gh workflow enable "Enable Pack Build and Test" --repo mamercad/stackstorm-exchange-ci; echo $?
could not find any workflows named Enable Pack Build and Test
1
```

cognifloyd:
Doh! Then I guess we do need an is-not-active check. Bummer.

Hmm. That check is baked into the `gh` client, so maybe the API won't have the same gotcha. So, we could also try using `gh api .../enable` instead of `gh workflow enable`.

I'm fine either way. Wdyt?

mamercad:
> Doh! Then I guess we do need an is-not-active check. Bummer.
> 
> Hmm. That check is baked into the `gh` client, so maybe the API won't have the same gotcha. So, we could also try using `gh api .../enable` instead of `gh workflow enable`.
> 
> I'm fine either way. Wdyt?

Yeah, that's a good thought, I'll iterate in the PR.

PR #138, opened 2022-12-18
CI: Add Python 3.8 to testing

mamercad:
Here's an [example](https://github.com/mamercad/stackstorm-jira/actions/runs/3725233753) running on my forks and branches; relates to [this](https://github.com/StackStorm-Exchange/stackstorm-jira/pull/49).

### DIFF
--- a/.github/workflows/pack-build_and_test.yaml
+++ b/.github/workflows/pack-build_and_test.yaml
@@ -35,7 +35,7 @@ jobs:
   build_and_test:
     runs-on: ubuntu-20.04
     # When parent workflow is named "Build and Test" this shows up as:
-    #   "Build and Test / Python 3.6"
+    #   "Build and Test / Python 3.X"
     name: 'Python ${{ matrix.python-version-short }}'
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


.github/workflows/pack-build_and_test.yaml
```diff
@@ -35,20 +35,24 @@ jobs:
...
     name: 'Python ${{ matrix.python-version-short }}'
     strategy:
+      fail-fast: false
+      max-parallel: 2
       matrix:
         include:
           - python-version-short: "3.6"
             python-version: 3.6.13
+          - python-version-short: "3.8"
+            python-version: 3.8.10
```
armab, line 48: Nice! :+1: to include 3.8 in the matrix.

@cognifloyd  Is there anything else we need to handle for this?

cognifloyd: There are a few repos that will also need an update I believe. I'll look at which when I'm at my computer later today.

cognifloyd:
Superseded by #142 and #145. Sorry I forgot about this and never looked into it.

PR #139, opened 2022-12-20
Conditionally self-enable "Pack Build and Test" workflow

mamercad:
Might address [this](https://github.com/StackStorm-Exchange/ci/pull/137#issuecomment-1358864596) in #137.

### DIFF
--- a/.github/workflows/enable-pack-build-and-test.yaml
+++ b/.github/workflows/enable-pack-build-and-test.yaml
@@ -59,9 +59,11 @@ jobs:
           fetch-depth: 1
 
       # TODO: Make sure this actually constitutes "repository activity"
-      - name: Set this Workflow to active (prevent 60d disabling)
+      - name: Conditionally set this Workflow to active (attempt to prevent 60d disabling of itself)
         shell: bash
         run:
+          # NOTE: "gh workflow list" does not currently support json output
+          # Relates to: https://github.com/cli/cli/issues/5025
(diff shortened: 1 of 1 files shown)


.github/workflows/enable-pack-build-and-test.yaml
```diff
@@ -59,10 +59,14 @@ jobs:
           fetch-depth: 1
 
       # TODO: Make sure this actually constitutes "repository activity"
-      - name: Set this Workflow to active (prevent 60d disabling)
+      - name: Conditionally set this Workflow to active (attempt to prevent 60d disabling of itself)
```
mamercad, line 62: Hrm, this might only restore this workflow's state (if it's disabled) when triggered by `on.workflow_dispatch`...

cognifloyd: workflow_dispatch is still a good escape hatch if all things automated fail for whatever set of reasons (like we're still figuring out what set of automation is most reliable).

cognifloyd (approved):

cognifloyd:
I merged, but missed a couple issues, so I pushed a couple of follow-up commits to fix this step.

That got us to this point:
https://github.com/StackStorm-Exchange/ci/actions/runs/3746070101/jobs/6361092828#step:6:59

Cloning repos can be occassionally flaky. For the index, that's not an issue because we do it so often, that waiting a few more minutes for the next run doesn't really matter. For this workflow, I would like to avoid cloning the pack repos somehow, so that this can be more reliable.

PR #140, opened 2022-12-20
CI: Pack metadata validation

### LINKED ISSUE #134: Add CI Checks: Changelog + Pack version update
https://github.com/StackStorm/st2/pull/5564 added Changelog checker for stackstorm/st2 based on https://github.com/marketplace/actions/changelog-checker which works for PR forks.

90% of the PR reviews in Exchange were about asking to add the missing changelog and version.
We'll need GH Actions automation for StackStorm-Exchange notifying users to update 2 things via PR checks/messages:
- `Changelog.md`
- Version in `pack.yaml`

mamercad:
Ensure `CHANGES.md` or `CHANGELOG.md` and `pack.yaml` has been updated; in addition, ensure that the proposed pack version in `pack.yaml` is higher.

Here are some example runs in my forks and branches:

- [`CHANGES.md` or `CHANGELOG.md` not updated](https://github.com/mamercad/stackstorm-test-pack/actions/runs/3742313106)
- [`pack.yaml` not updated](https://github.com/mamercad/stackstorm-test-pack/actions/runs/3742331784)
- [Proposed pack version equal](https://github.com/mamercad/stackstorm-test-pack/actions/runs/3742342955)
- [Proposed pack version lower](https://github.com/mamercad/stackstorm-test-pack/actions/runs/3742359747)
- [All checks passing](https://github.com/mamercad/stackstorm-test-pack/actions/runs/3742361691)

Fixes #134.

### DIFF
--- a/.github/workflows/pack-build_and_test.yaml
+++ b/.github/workflows/pack-build_and_test.yaml
@@ -49,22 +49,22 @@ jobs:
           st2-branch: ${{ inputs.st2-branch }}
           lint-configs-branch: ${{ inputs.lint-configs-branch }}
 
-      # - name: Install APT Dependencies
-      #   uses: StackStorm-Exchange/ci/.github/actions/apt-dependencies@master
-      #   with:
-      #     cache-version: ${{ inputs.apt-cache-version }}
+      - name: Install APT Dependencies
+        uses: StackStorm-Exchange/ci/.github/actions/apt-dependencies@master
+        with:
+          cache-version: ${{ inputs.apt-cache-version }}
 
-      # - name: Install Python Dependencies
-      #   uses: StackStorm-Exchange/ci/.github/actions/py-dependencies@master
-      #   with:
-      #     cache-version: ${{ inputs.py-cache-version }}
-      #     python-version: ${{ matrix.python-version }}
+      - name: Install Python Dependencies
+        uses: StackStorm-Exchange/ci/.github/actions/py-dependencies@master
+        with:
+          cache-version: ${{ inputs.py-cache-version }}
+          python-version: ${{ matrix.python-version }}
 
-      # - name: Run pack tests
-      #   uses: StackStorm-Exchange/ci/.github/actions/test@master
-      #   with:
-      #     enable-common-libs: ${{ inputs.enable-common-libs }}
-      #     force-check-all-files: ${{ inputs.force-check-all-files }}
+      - name: Run pack tests
+        uses: StackStorm-Exchange/ci/.github/actions/test@master
+        with:
+          enable-common-libs: ${{ inputs.enable-common-libs }}
+          force-check-all-files: ${{ inputs.force-check-all-files }}
 
       - name: Ensure CHANGES.md and pack.yaml changed (and Pack version increased)
         uses: mamercad/stackstorm-exchange-ci/.github/actions/pack-meta@changelog-and-pack-checking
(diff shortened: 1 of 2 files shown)


cognifloyd (commented):
Are there cases where the workflow should not fail because of this?

We need it for Pull Requests, but I'm concerned about how this night behave when it runs:
- on merge to master
- on the weekly schedule

And, are there cases where we might need to do exchange-wide updates for CI, or something that does not affect what gets installed with the pack? Where kind of escape hatch would work here?

If we can skip this if the workflow is not running for a pull request, then that would probably resolve all my concerns as exchange-wide updates get pushed to master to minimize the notification spam. Is there a good way to do that?

mamercad:
> If we can skip this if the workflow is not running for a pull request, then that would probably resolve all my concerns as exchange-wide updates get pushed to master to minimize the notification spam. Is there a good way to do that?

Ah yeah, I missed the fact that it runs on a weekly schedule, and, it run during release. We should be able to skip it unless it's a PR, let me work on it.

.github/workflows/pack-build_and_test.yaml
```diff
@@ -66,6 +66,10 @@ jobs:
           enable-common-libs: ${{ inputs.enable-common-libs }}
           force-check-all-files: ${{ inputs.force-check-all-files }}
 
+      - name: Ensure CHANGES.md and pack.yaml changed (and Pack version increased)
+        if: ${{ github.event_name == 'pull_request' }}
+        uses: StackStorm-Exchange/ci/.github/actions/pack-meta@master
```
armab, line 71: It would be nice to have a separate named checks for those items:
- Increment pack version in `pack.yaml`
- Add a changelog in `CHANGES.md`

So they show up in Github CI status for every PR independently of each other, also are self-describing and run outside of the pack tests.

armab: Screenshot 2022-12-20 at 23-52-13 Pull requests · StackStorm_st2
in `stackstorm/st2` is a good example

mamercad: No problem, is there a "pack schema" anywhere? What I was wondering about when working on this was, is it `CHANGES.md` or `CHANGELOG.md` or `CHANGELOG.rst` or `CHANGE*`?

armab: For StackStorm-Exchange it should be always `CHANGES.md` and is standardized around the Exchange repos:
https://github.com/StackStorm-Exchange/stackstorm-jira/blob/master/CHANGES.md

I think making sure there's a diff for these two files in the PR would be sufficient, without going deep into the pack schema ;) 
Most of the times we ask people to bump the pack version in the pack.yaml and update the CHANGES.md file while reviewing the PRs in Exchange. https://github.com/StackStorm-Exchange/stackstorm-rabbitmq/pull/20#pullrequestreview-1221809999

Doing that via automation as an additional CI status check would be super cool already!

mamercad: Is there anything left to do in this PR?

cognifloyd (approved):
Looks good to me. Once we have some experience with this, we can revise (more workflows/jobs/naming/etc) if needed.

mamercad:
> Looks good to me. Once we have some experience with this, we can revise (more workflows/jobs/naming/etc) if needed.

Continual refinement, as always 😄

PR #141, opened 2022-12-21
Enable pack "Build and Test" workflow without cloning && enable using fine-grained PAT

cognifloyd:
This simplifies the workflow based on my comment here: https://github.com/StackStorm-Exchange/ci/pull/139#issuecomment-1360800869

> Cloning repos can be occassionally flaky. For the index, that's not an issue because we do it so often, that waiting a few more minutes for the next run doesn't really matter. For this workflow, I would like to avoid cloning the pack repos somehow, so that this can be more reliable.

And it switches from `gh workflow` to `gh api` so that we can use fine-grained PATs.

### DIFF
--- a/.github/workflows/enable-pack-build-and-test.yaml
+++ b/.github/workflows/enable-pack-build-and-test.yaml
@@ -40,7 +40,6 @@ jobs:
         run: |
           echo "CI_DIR=${{ github.workspace }}/ci" >> ${GITHUB_ENV}
           echo "TOOLS_DIR=${{ github.workspace }}/exchange-tools" >> ${GITHUB_ENV}
-          echo "PACKS_PATH=${{ github.workspace }}/packs" >> ${GITHUB_ENV}
 
       - name: Checkout ci repo
         uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


cognifloyd:
@mamercad and @armab wdyt?

cognifloyd:
The last commit I added should, I hope, allow us to use a fine-grained PAT with this. I ran into some permissions issues with the bot token I created:
https://github.com/StackStorm-Exchange/ci/actions/runs/3746070101/jobs/6361092828#step:6:69
```
  failed to enable workflow: HTTP 403: Must have admin rights to Repository. (https://api.github.com/repos/StackStorm-Exchange/stackstorm-acos/actions/workflows/16569224/enable)
```

But, as @mamercad noticed, graphql doesn't work with a fine-grained PAT, and `gh workflow` apparently is using graphql. :facepalm:
https://github.com/StackStorm-Exchange/ci/actions/runs/3746417874/jobs/6361741025#step:6:59
```
gh: Personal access tokens with fine grained access do not support the GraphQL API (HTTP 401)
```

`gh api` explicitly does not use graphql. So, I'm hoping this will resolve those issues.

The fine-grained PAT must have an expiration date, so we will have to manually rotate this secret occasionally.

.github/workflows/enable-pack-build-and-test.yaml
```diff
@@ -61,60 +60,35 @@ jobs:
...
         shell: bash
+        env:
+          WORKFLOW_PATH: repos/${{ github.repository }}/actions/workflows/enable-pack-build-and-test.yaml
         run: |
           # NOTE: "gh workflow list" does not currently support json output
           # Relates to: https://github.com/cli/cli/issues/5025
-          if [[ $(gh api --method GET "repos/${{ github.repository }}/actions/workflows/enable-pack-build-and-test.yaml" --jq .state) != "active" ]]; then
-            gh workflow enable "${{ github.workflow }}" --repo "${{ github.repository }}"
+          if [[ $(gh api --method GET "${WORKFLOW_PATH}" --jq .state) != "active" ]]; then
+            gh api --method PUT "${WORKFLOW_PATH}/enable"
```
cognifloyd, line 69: Since we're using `gh api` here, we don't actually have to check to see if `!= "active"`, because the API does not care what the current state is, it just enables it. The restriction was in the `gh workflow` tool.

So, wdyt? Drop the condition to reduce the number of API calls?

mamercad: > Drop the condition to reduce the number of API calls?

Yeah, might as well.

mamercad:
> @mamercad and @armab wdyt?

Looks good to me!

armab (approved):
I like the enhancement with simplified approach :+1:

PR #143, opened 2023-10-23
update default python version for packs to be a list

sravs-dev:


armab (commented):
@sravs-dev  Thanks for the PR!

Let's verify if it works in a pack CI.
You can open a PR in some Exchange pack repo. Let's pick up `Github` pack, the workflow is here: https://github.com/StackStorm-Exchange/stackstorm-github/blob/master/.github/workflows/build_and_test.yaml#L13
You can open a PR changing the ref to your fork with a branch.

That should use this GH Workflow and we can verify if it works OK.

Please also sign CLA.

.github/workflows/pack-build_and_test.yaml
```diff
@@ -14,7 +14,7 @@ on:
       python-version:
         required: false
         type: string
-        default: "3.6"
+        default: ["3.6","3.8","3.9"]
```
cognifloyd, line 17: Changing the type from string to array here would mean we have to change the matrix definition below. There, we are using `matrix.include` to add one entry for the indicated python version.

The docs about `matrix.include` are here.: https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#expanding-or-adding-matrix-configurations

cognifloyd (changes requested):

sravs-dev:
@armab I have signed the CLA, but it still shown as pending. Could you please help?

sravs-dev:
@armab opened a new [PR](https://github.com/StackStorm-Exchange/ci/pull/145) to resolve commit log issues. Closing this one

PR #146, opened 2023-12-06
replace deprecated `::set-output` w/ `${GITHUB_OUTPUT}`

cognifloyd:
Github deprecated using the `::set-output` command in workflows/actions. Instead, we are supposed to use a environment file at `${GITHUB_OUTPUT}`.

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


cognifloyd:
Ugh. CircleCI needs to be released with a GHA workflow in this repo.

armab (approved):

PR #147, opened 2024-01-16
Update github actions to use py3.10/3.11 for Ubuntu 22.04 support

nzlosh:


.github/workflows/pack-build_and_test.yaml
```diff
@@ -14,7 +14,7 @@ on:
       python-version:
         required: false
         type: string
-        default: '["3.6","3.8","3.9"]'
+        default: '["3.8", "3.9", "3.10", "3.11"]'
```
amanda11, line 17: Line 42 will need updating which says what the job output will be

amanda11 (commented):
Best to amend the comment that references 3.6 as well

jpavlav:
@mamercad Can we get another 👍 on this? It looks like it should do the trick and it's blocking a PR downstream that uses this repository.

mamercad (approved):
🚢

cognifloyd (approved):

PR #150, opened 2024-04-18
Drop python 3.6

cognifloyd:
Now that pip has been updated in the primary st2 repo, index updates are failing because that pip version is not available in py3.6


amanda11 (approved):

amanda11 (commented):
Do you also need to change the .circleci to have a run on python3.9 instead of python3.6?

cognifloyd:
Updated 
- CI repo .circleci config to use 3.9
- pack bootstrap .circleci config to use 3.8

Deleted an unused old copy of circleci config. 

Circle CI is now running. The pylint failures are around using f-strings and other upgrades, so we can deal with that in a different PR.

PR #151, opened 2024-10-11
Pin pip, setuptools and virtualenv modules to st2 project.

nzlosh:


nzlosh:
Pushed linting fixes so CircleCI pases.

cognifloyd (approved):

